PR #5, opened 2018-06-11
Added support for Travis CI

mvanbeusekom:

Feature

Currently there is no continuous integration in place.

Implemented continuous integration using Travis CI

Nope

n/a

More information on implementing Travis CI with Flutter can be found here:

- https://medium.com/flutter-io/test-flutter-apps-on-travis-3fd5142ecd8c

- [x] All projects build

### DIFF
--- a/test/geolocator_test.dart
+++ b/test/geolocator_test.dart
@@ -1,0 +1,47 @@
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+import 'package:mockito/mockito.dart';
+import 'package:flutter_geolocator/flutter_geolocator.dart';
+import 'package:flutter_geolocator/models/position.dart';
+import 'package:test/test.dart';
+
+void main() {
+  MockMethodChannel _methodChannel;
+  MockEventChannel _eventChannel;
+  FlutterGeolocator _geolocator;
+
+  const Map<String, double> _mockPosition = const <String, double> {
+    'latitude' : 52.561270, 
+    'longitude' : 5.639382,
+    'altitude' : 3000.0,
+    'accuracy' : 0.0,
+    'heading' : 0.0,
+    'speed' : 0.0,
+    'speed_accuracy' : 0.0
+  };
+
+  setUp(() {
+    _methodChannel = new MockMethodChannel();
+    _eventChannel = new MockEventChannel();
+    _geolocator = new FlutterGeolocator.private(_methodChannel, _eventChannel);
+  });
+
+  test('position', () async {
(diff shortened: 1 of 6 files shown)


test/geolocator_test.dart
```diff
@@ -0,0 +1,47 @@
...
+    expect(position.altitude, _mockPosition['altitude']);
+    expect(position.accuracy, _mockPosition['accuracy']);
+    expect(position.heading, _mockPosition['heading']);
+    expect(position.speed, _mockPosition['speed']);
+    expect(position.speedAccuracy, _mockPosition['speed_accuracy']);
+  });
+}
+
+class MockMethodChannel extends Mock implements MethodChannel {}
+class MockEventChannel extends Mock implements EventChannel {}
```
mravn-google: We often don't mock channel objects since they are just lightweight/stateless wrappers around their name and codec. So our tests generally use real channel instances. `MethodChannel` has [built-in mocking support](https://docs.flutter.io/flutter/services/MethodChannel/setMockMethodCallHandler.html), and incoming messages (platform -> Dart) such as events can be simulated using [`BinaryMessages.handlePlatformMessage`](https://docs.flutter.io/flutter/services/BinaryMessages/handlePlatformMessage.html) and the codec of the (real) `EventChannel`.

test/geolocator_test.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  setUp(() {
+    _methodChannel = new MockMethodChannel();
+    _eventChannel = new MockEventChannel();
+    _geolocator = new FlutterGeolocator.private(_methodChannel, _eventChannel);
+  });
+
+  test('position', () async {
+    when(_methodChannel.invokeMethod('getPosition'))
+      .thenReturn(new Future<Map<String, double>>.value(_mockPosition));
```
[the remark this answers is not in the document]
mvanbeusekom: Thanks, this indeed makes things easier. Will make the necessary adjustments.

test/geolocator_test.dart
mvanbeusekom: I must admit that I used the (Battery)[https://github.com/flutter/plugins/tree/master/packages/battery] plugin to base my tests on. I made the mistake that mocking the channel objects would be a good way to test the conversion of the returned `Map<String, double>` into an instance of `Position` class. However this is of course of no concern of the `getPosition` method but should (and is) be tested on the `Position` class it self. 

So to make a long story short: thanks I will also update the code here ;)

PR #10, opened 2018-06-12
Make sure state is cleaned up correctly

mvanbeusekom:

Bug fix

The flutter analyser complains about two issues in the example app:
1. The value of the field '_positionSubscription' isn't used.
2. Cancel instances of `dart.async.StreamSubscription`.

This PR solves both problems by using the field `_positionSubscription` to cancel the `StreamSubscription` when the state is being disposed.

No

Run the example in a simulator

- [dispose method](https://docs.flutter.io/flutter/widgets/State/dispose.html)

- [x] All projects build

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -66,4 +66,14 @@ class _MyAppState extends State<MyApp> {
       ),
     );
   }
+
+  @override
+  void dispose() {
+    if(_positionSubscription != null) {
+      _positionSubscription.cancel();
+      _positionSubscription = null;
+    }
+
+    super.dispose();
+  }
 }


example/lib/main.dart
```diff
@@ -66,4 +66,14 @@ class _MyAppState extends State<MyApp> {
       ),
     );
   }
+
+  @override
+  void dispose() {
+    if(_positionSubscription != null) {
```
mravn-google: It is considered a best practice in Dart land to let [`dartfmt`](https://github.com/dart-lang/dart_style#readme) format all your code. In this case it would add a space after `if`. Flutter tooling packs `dartfmt` so you can do
```
$ cd myFlutterProject
$ flutter format .
```
This will rewrite any Dart file in your project that isn't already correctly formatted.

PR #12, opened 2018-06-20
Implemented suggestions from Mikkel

mvanbeusekom:
Implemented all suggestions made by Mikkel after he conducted a code review.

The only suggestion I skipped for now is the remark regarding the mocking of the `MethodChannel` and `EventChannel` classes. I will have to do a little bit more research on how to correctly implement this.

### DIFF
--- a/android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
+++ b/android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
@@ -55,7 +55,7 @@ public class GeolocatorPlugin implements MethodCallHandler, EventChannel.StreamH
   private EventChannel.EventSink mEventSink;
   private Result mResult;
 
-  private GeolocatorPlugin(PluginRegistry.Registrar registrar){
+  private GeolocatorPlugin(PluginRegistry.Registrar registrar) {
     this.mRegistrar = registrar;
 
     mFusedLocationClient = LocationServices.getFusedLocationProviderClient(mRegistrar.activity());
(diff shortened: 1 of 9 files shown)


test/geolocator_test.dart
```diff
@@ -112,25 +113,27 @@ void main() {
     });
 
     test('Receive position changes', () async {
-      final StreamQueue<Position> queue = new StreamQueue<Position>(_geolocator.onPositionChanged);
+      final StreamQueue<Position> queue =
+          new StreamQueue<Position>(_geolocator.onPositionChanged);
```
mravn-google: This test checks that `_geolocator.onPositionChanged` is the stream returned by `_eventChannel.receiveBroadcastStream()` and that `Position.toMap` works for the test data. But otherwise it seems to be mostly testing that `StreamController` and `StreamQueue` work.

If one wanted to go a step further, one could include position and error serialization/deserialization to/from the binary message form in the testing by using `BinaryMessages` to synthesize incoming messages from the platform. That can be done by replacing use of the `StreamController` with something like
```dart
void simulateIncomingEvent(Map<String, double> position) {
  BinaryMesssages.handlePlatformMessage(
    _eventChannel.name,
    _eventChannel.codec.encodeSuccessEnvelope(position),
    (_) {},
  );
}

void simulateIncomingError(String code, String message) {
  BinaryMessages.handlePlatformMessage(
    _eventChannel.name,
    _eventChannel.codec.encodeErrorEnvelope(code, message, null),
    (_) {},
  );
}
```
Of course, this requires using a real `EventChannel` object as part of the test, not just a mock. In terms of regression testing, we would guard against later including values in `Position.toMap` that cannot be handled by the channel codec (e.g. time stamps).

android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
```diff
@@ -177,22 +177,33 @@ private void createPermissionsResultListener() {
...
+                for (Result result : mResultHandles) {
+                  result.error("PERMISSION_DENIED_NEVER_ASK",
+                      "Access to location data denied. To allow access to location services enable them in the device settings.",
+                      null);
+                }
+
+                mResultHandles.clear();
               } else if (mEventSink != null) {
-                mEventSink.error("PERMISSION_DENIED_NEVER_ASK", "Access to location data denied. To allow access to location services enable them in the device settings.", null);
+                mEventSink.error("PERMISSION_DENIED_NEVER_ASK",
```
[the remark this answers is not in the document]
mvanbeusekom: You are correct, it is absolutely possible to have non-zero result handles and an event sink. I have updated the code to fix this.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
```diff
@@ -229,13 +242,18 @@ public void onSuccess(Location location) {
...
+            mResultHandles.clear();
           }
         } else {
-          if (mResult != null) {
-            mResult.error("ERROR", "Failed to get location.", null);
+          if (mResultHandles.size() > 0) {
+            for (Result result : mResultHandles) {
+              result.error("ERROR", "Failed to get location.", null);
+            }
+            mResultHandles.clear();
```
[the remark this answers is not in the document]
mvanbeusekom: I was thinking the same last night but couldn't figure out how to implement Java lambda's at the time. Having a fresh look this morning I figured it out (got stuck on where to configure the target compatibility version for java so it support java 1.8 features). 

Thanks for pointing it out and giving me that little bit of extra motivation to look into it again ;)

android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
```diff
@@ -40,6 +41,11 @@
  */
 public class GeolocatorPlugin implements MethodCallHandler, EventChannel.StreamHandler {
 
+  @FunctionalInterface
+  private interface ResultLambdaExpression {
+    void ProcessResult(Result result);
```
mravn-google: [nit] PascalCase in Java method name? Hurts my eyes 😭

I don't suppose `java.util.function.Consumer` is available across all relevant Android SDKs? If it is, you could replace this by `Consumer<Result>`.

mvanbeusekom: hahaha sorry really have an extensive C# background, some habits die hard ;)

martijn00:
@mvanbeusekom can you do a rebase on this branch?

android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
mvanbeusekom: Just quickly checked, `java.util.function.Consumer` is only available from Android API 24 and up. Since we are targeting minimal Android API 16, I think it is best to leave it as is (of course with the renamed method)

mravn-google: Roger that. You seem to have left an import of the `Consumer` interface in the code (line 34).

mvanbeusekom: Check

PR #25, opened 2018-07-03
Support to supply desired accuracy

### LINKED ISSUE #1: Configure location refresh interval
Currently the interval on which the location will be read is hard coded. It would be nice if we can update this so that the developer can change the interval at run-time.  

The use-case here would be to allow developers to dial back or increase the frequency depending on the state off the App (e.g. background vs. foreground).

mvanbeusekom:

Feature (resolves #1 )

Currently it is not possible to supply a desired accuracy, possible resulting in high battery use.

It is now possible to supply a desired accuracy when requesting the position.

Yes (see  [CHANGELOG.md](https://github.com/BaseflowIT/flutter-geolocator/blob/feature/configuration/CHANGELOG.md))

Run unit-tests: `flutter test` (or check results from Travis-CI)

Resolves #1 

- [x] All projects build


android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
```diff
@@ -31,230 +13,107 @@
...
-            && permissions[0].equals(Manifest.permission.ACCESS_FINE_LOCATION)) {
-          if (grantResults[0] == PackageManager.PERMISSION_GRANTED) {
-            acquirePosition();
-          } else {
-            if (!shouldShowRequestPermissionRationale()) {
-                SignalResultHandles(result -> result.error("PERMISSION_DENIED_NEVER_ASK",
-                      "Access to location data denied. To allow access to location services enable them in the device settings.",
-                      null));
+    public void onCompletion(UUID taskID) {
+        Iterator<Map.Entry<UUID, Result>> it = mResultHandles.entrySet().iterator();
```
mravn-google: `mResultHandles.remove(taskID);`?

android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
```diff
@@ -31,230 +13,107 @@
...
-            } else {
-              SignalResultHandles(result -> result.error("PERMISSION_DENIED", "Access to location data denied", null));
+        while (it.hasNext()) {
+            Map.Entry<UUID, Result> entry = it.next();
 
-              if (mEventSink != null) {
-                mEventSink.error("PERMISSION_DENIED", "Access to location data denied", null);
-                mEventSink = null;
-              }
+            if (taskID == entry.getKey()) {
```
mravn-google: Should use `.equals` instead. Otherwise you are comparing object references.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/GeolocatorPlugin.java
```diff
@@ -31,230 +13,107 @@
...
-          Map<String, Double> position = LocationMapper.toHashMap(location);
+    private static GeolocationAccuracy parseAccuracy(Object o) {
+        GeolocationAccuracy accuracy = GeolocationAccuracy.Medium;
 
-          SignalResultHandles(result -> result.success(position));
-        } else {
-          SignalResultHandles((result) -> result.error("ERROR", "Failed to get location.", null));
+        try {
+            int index = (Integer) o;
+            if(index > 0 && index < GeolocationAccuracy.values().length) {
```
mravn-google: That should probably be `if(index >= 0 && ...`

PR #34, opened 2018-07-09
Add templates for issues and improve readme

martijn00:

Docs

 No issue templates

Yay templates!

No

Click on create new issue

- [x] All projects build

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ master  | [![Build Status](https://travis-ci.com/BaseflowIT/flutter-geolocator.s
 
 ## Features
 
-* Get current location one-time
+* Get the current location of the device
 * Get continuous location updates
 
 ## Usage
@@ -98,7 +98,7 @@ Please file any issues, bugs or feature request as an issue on our [GitHub](http
 
 ## Want to contribute
 
-I you would like to contribute to the plugin (e.g. by improving the documentation, solving a bug or adding a cool new feature), please carefully review our [contribution guide](CONTRIBUTING.md) and send us your [pull request](https://github.com/BaseflowIT/flutter-geolocator/pulls).
+If you would like to contribute to the plugin (e.g. by improving the documentation, solving a bug or adding a cool new feature), please carefully review our [contribution guide](CONTRIBUTING.md) and send us your [pull request](https://github.com/BaseflowIT/flutter-geolocator/pulls).
 
 ## Author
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -22,6 +23,34 @@ dependencies:
 
 > **NOTE:** There's a known issue with integrating plugins that use Swift into a Flutter project created with the Objective-C template. See issue [Flutter#16049](https://github.com/flutter/flutter/issues/16049) for help on integration.
 
+## Example
```
mvanbeusekom: I think we should start with the `Permissions` topic. This will make sure people will read it before jumping into the example code and possible bother us with questions regarding why their code isn't working.

README.md
```diff
@@ -97,4 +98,8 @@ Please file any issues, bugs or feature request as an issue on our [GitHub](http
 
 ## Want to contribute
 
-I you would like to contribute to the plugin (e.g. by improving the documentation, solving a bug or adding a cool new feature), please carefully review our [contribution guide](CONTRIBUTING.md) and send us your [pull request](https://github.com/BaseflowIT/flutter-geolocator/pulls).
\ No newline at end of file
+I you would like to contribute to the plugin (e.g. by improving the documentation, solving a bug or adding a cool new feature), please carefully review our [contribution guide](CONTRIBUTING.md) and send us your [pull request](https://github.com/BaseflowIT/flutter-geolocator/pulls).
```
mvanbeusekom: I see that the original author (me) made a small type here `I you would like` should be `If you would like`. Maybe you can correct it?

README.md
```diff
@@ -8,7 +8,8 @@ develop | [![Build Status](https://travis-ci.com/BaseflowIT/flutter-geolocator.s
 master  | [![Build Status](https://travis-ci.com/BaseflowIT/flutter-geolocator.svg?branch=master)](https://travis-ci.com/BaseflowIT/flutter-geolocator)
 
 ## Features
-* Get one-shot location
+
+* Get current location one-time
```
mvanbeusekom: I would remove the word `one-time` as it suggests you can only get the current location once (which is not the case). Better would be `Get the current location of the device`

README.md
martijn00: I was thinking we should start with the example because people always want to see some real code. They will look further into the permissions once they use it.

mvanbeusekom: Hmm I consider the permissions also a bit as example code (it is the first step you need to take). @spockfish what do you think?

PR #35, opened 2018-07-10
Added geocoding features

### LINKED ISSUE #24: Translate address to coördinates
Add support to translate a given address into a list of coördinates

mvanbeusekom:

Feature

The current plugin doesn't support translation of an address into geocoordinates and vice versa.

The new version does support translation of an address into geocoordinates and vice versa.

No

Run unit-tests and the example application

- Documentation is added in the [README.md](README.md) file;
- Fixes #24

- [x] All projects build
- [X] Follows style guide lines ([code style guide](https://github.com/BaseflowIT/flutter-geolocator/blob/develop/CONTRIBUTING.md))


lib/geolocator.dart
```diff
@@ -39,6 +40,28 @@ class Geolocator {
...
+    var placemarks = await _methodChannel.invokeMethod('toPlacemark', address);
+    return Placemark.fromMaps(placemarks);
+  }
+
+  /// Returns a list of [Placemark] instances found for the supplied address.
+  ///
+  /// In most situations the returned list should only contain one entry.
+  /// However in some situations where the supplied address could not be
+  /// resolved into a single [Placemark], multiple [Placemark] instances may be returned.
+  Future<List<Placemark>> fromPlacemark(
```
mravn-google: Strangely named method. `toPlacemark` turns an address into a (list of) `Placemark`. But `fromPlacemark` does not convert a `Placemark` back to an address.

lib/models/placemark.dart
```diff
@@ -0,0 +1,75 @@
...
+  final String thoroughfare;
+
+  /// Additional street address information for the placemark.
+  final String subThoroughfare;
+
+  /// The geocoordinates associated with the placemark.
+  final Position position;
+
+  /// Converts a list of [Map] instances to a list of [Placemark] instances.
+  static List<dynamic> fromMaps(dynamic message) {
```
mravn-google: The return type should be `List<Placemark>`. This method should probably be private.

lib/models/placemark.dart
```diff
@@ -0,0 +1,75 @@
...
+    var list = message.map<Placemark>(fromMap).toList();
+    return list;
+  }
+
+  /// Converts the supplied [Map] to an instance of the [Placemark] class.
+  static Placemark fromMap(dynamic message) {
+    final Map<dynamic, dynamic> placemarkMap = message;
```
mravn-google: Cf. earlier remark on suppressing type errors in channel communication. This code throws a nice runtime exception, if the Java/Swift side does not return a map. It is a bit inconsistent that argument errors are checked and suppressed while return value errors are not.

lib/models/position.dart
```diff
@@ -43,6 +43,11 @@ class Position {
   /// The speedAccuracy is not available on all devices. In these cases the value is 0.0.
   final double speedAccuracy;
 
+  static List<dynamic> fromMaps(dynamic message) {
```
mravn-google: I'd keep platform message conversion methods private, as they are an implementation detail.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/Result.java
```diff
@@ -0,0 +1,35 @@
+package com.baseflow.flutter.plugin.geolocator.data;
+
+import io.flutter.plugin.common.EventChannel;
+import io.flutter.plugin.common.MethodChannel;
+
+public class Result {
```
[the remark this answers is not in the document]
mvanbeusekom: @mravn-google would it be an acceptable solution when I split the results up in a `MethodResult` and a `EventResult` class that both implement a `ResultInterface`?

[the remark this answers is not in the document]
mravn-google: That might help---but only if you use the subtypes in `Task` implementations. Perhaps `Task<R extends ResultInterface>`? Dunno. May be overkill.

Your current solution is already acceptable. I just wanted to highlight that even though both `MethodChannel.Result` and `EventChannel.EventSink` have `success` and `error` methods, they do have different semantics. Hiding that may never actually lead you to a bug, and so you probably don't want to overengineer it.

lib/models/position.dart
mvanbeusekom: @mravn-google probably a stupid question but I just don't see it: how would I handle this when I need to call the `fromMaps` (or `fromMap`) methods in from the `Geolocator` class?

mravn-google: You can make all of your implementation classes parts of a single Dart library. Then all the classes can access each other's private members, but client code cannot.

There are two ways to do that. Either you put all the classes into the same Dart file, or you keep them in separate files, linked by `library`, `part`, and `part of` declarations. See e.g. [the firebase_database plugin](https://github.com/flutter/plugins/tree/master/packages/firebase_database/lib) for an example of how to set up the latter.

mvanbeusekom: Got it, works like a charm! Thank you for the explanation

PR #46, opened 2018-07-18
Implement API to calculate distance between two positions

### LINKED ISSUE #23: Calculate distance between two locations
Nice extra feature to have would be the option to calculate the distance between two points.

mvanbeusekom:

Feature

The current plugin doesn't support calculating the distance between two points.

This PR offers a new `distanceBetween` API method to calculate the distance (in meters) between two points.

No

Run the example app

Resolves #23 

- [x] All projects build

### DIFF
--- a/android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/Coordinate.java
+++ b/android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/Coordinate.java
@@ -1,10 +1,10 @@
 package com.baseflow.flutter.plugin.geolocator.data;
 
-public class Coordinates {
+public class Coordinate {
     private final Double mLatitude;
     private final Double mLongitude;
 
-    public Coordinates(Double latitude, Double longitude) {
+    public Coordinate(Double latitude, Double longitude) {
         mLatitude = latitude;
         mLongitude = longitude;
     }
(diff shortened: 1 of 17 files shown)


mvanbeusekom:
Yes good spot, forgot about updating the readme file

android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/Coordinate.java
```diff
@@ -1,10 +1,10 @@
 package com.baseflow.flutter.plugin.geolocator.data;
 
-public class Coordinates {
+public class Coordinate {
```
mravn-google: Sorry, missed this earlier. Do you really want to allow `null` latitude or longitude? If not, I suggest making the fields (and getters) have type `double` so that illegal state is not representable.

[nit] It is not clear to my why the fields are encapsulated in such a trivial value class. Is there any reason to not just do
```java
public final class Coordinate {
  public final double latitude;
  public final double longitude;

  public Coordinate(double latitude, double longitude) {
    this.latitude = latitude;
    this.longitude = longitude;
  }
}
```

example/lib/widgets/current_location_widget.dart
```diff
@@ -0,0 +1,65 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:geolocator/geolocator.dart';
+import 'package:geolocator/models/location_accuracy.dart';
+import 'package:geolocator/models/location_options.dart';
+import 'package:geolocator/models/position.dart';
+
+class CurrentLocationWidget extends StatefulWidget {
```
mravn-google: I think maybe this class could be much simplified by using the `StreamBuilder` widget. Then you don't have to worry about subscriptions and what-not.

PR #79, opened 2018-08-22
Set default distance filter value to 0m for FusedLocationProviderClient

audkar:

Bug fix

- Currently location changes reported only when distance changes by at least 1m when using `FusedLocationProviderClient`. 
- All location changes reported when using `LocationManager` (distance filter 0 by default).

All locations changes reported by default for both location providers.

No

- [x] All projects build

### DIFF
--- a/android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/LocationUpdatesUsingLocationServicesTask.java
+++ b/android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/LocationUpdatesUsingLocationServicesTask.java
@@ -83,14 +83,10 @@ class LocationUpdatesUsingLocationServicesTask extends LocationUsingLocationServ
     private LocationRequest createLocationRequest() {
         LocationRequest locationRequest = new LocationRequest();
 
-        float distanceFilter = mLocationOptions.distanceFilter > 0
-                ? mLocationOptions.distanceFilter
-                : 1;
-
         locationRequest
                 .setInterval(0)
                 .setFastestInterval(0)
-                .setSmallestDisplacement(distanceFilter);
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/LocationUpdatesUsingLocationServicesTask.java
```diff
@@ -85,7 +85,7 @@ private LocationRequest createLocationRequest() {
 
         float distanceFilter = mLocationOptions.distanceFilter > 0
                 ? mLocationOptions.distanceFilter
-                : 1;
+                : 0;
```
mvanbeusekom: I have explicitly set the distance filter to 1 meter because I had troubles receiving events when I left it to 0. This could very well be something I was doing wrong though.

Anyway if we want to set it by default to 0 we could just use the following code:

```
float distanceFilter = mLocationOptions.distanceFilter;
```

(so remove the inline if statement)

audkar: `mLocationOptions.distanceFilter` can be passed as negative number. But that probably would be devs problem :)))
`0` is default [value](https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest.html#setSmallestDisplacement(float)) so that should work in theory.

I am facing problem when GPS gets very accurate data and device is stationary then geolocator stops sending updates. 

Have you tested on real device with `0` value?

mvanbeusekom:
@audkar we just released the fix to Dart Packages. If you update to version 1.6.1 this problem should be solved.

Thanks again for contributing.

PR #88, opened 2018-08-30
[WIP] Add magnetometer for Android

martijn00:

Feature

No Magnetometer support

Added initial code for Android

No

Base for https://github.com/BaseflowIT/flutter-geolocator/issues/86

- [ ] All projects build

### DIFF
--- a/android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
+++ b/android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
@@ -1,0 +1,62 @@
+package com.baseflow.flutter.plugin.geolocator.tasks;
+
+import android.content.Context;
+import android.location.Address;
+import android.location.Geocoder;
+
+import com.baseflow.flutter.plugin.geolocator.data.AddressMapper;
+import com.baseflow.flutter.plugin.geolocator.data.Result;
+
+import java.io.IOException;
+import java.util.List;
+
+import io.flutter.plugin.common.PluginRegistry;
+
+class MagnetometerTask extends Task implements SensorEventListener {
+    private final Context mContext;
+
+    private SensorManager sensorManager;
+    Sensor mSensor; // Magnetic sensor returned by sensor manager
+
+    float x; // magnetometer x value
+    float y; // magnetometer y value
+    float z; // magnetometer z value
+    float magnitude; // magnetometer calculated magnitude
+
+    public MagnetometerTask(TaskContext context) {
+        super(context);
+
+        PluginRegistry.Registrar registrar = context.getRegistrar();
+        mContext = registrar.activity() != null ? registrar.activity() : registrar.activeContext();
+    }
+
+    SensorManager getSensorManager() {
+        Context context = mContext;
+        return (SensorManager) context.getSystemService(Context.SENSOR_SERVICE);
+    }
+
+    @Override
+    public void startTask() {
+
+        sensorManager = getSensorManager();
+        List<Sensor> list = sensorManager.getSensorList(Sensor.TYPE_MAGNETIC_FIELD);
+
+        // If found, then register as listener
+        if (list != null && list.size() > 0) {
+            mSensor = list.get(0);
+            sensorManager.registerListener(this, mSensor, SensorManager.SENSOR_DELAY_NORMAL);
+        }
+    }
+
+    public void onSensorChanged(SensorEvent event) {
+        this.x = event.values[0];
+        this.y = event.values[1];
+        this.z = event.values[2];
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
```diff
@@ -0,0 +1,62 @@
...
+    public MagnetometerTask(TaskContext context) {
+        super(context);
+
+        PluginRegistry.Registrar registrar = context.getRegistrar();
+        mContext = registrar.activity() != null ? registrar.activity() : registrar.activeContext();
+    }
+
+    SensorManager getSensorManager() {
+        Context context = mContext;
+        return (SensorManager) context.getSystemService(Context.SENSOR_SERVICE);
```
mvanbeusekom: You can safely initialize the `sensorManager` inside the constructor (see also example [here](https://developer.android.com/reference/android/hardware/SensorManager)).

This way we don't set unnecessarily overwrite the `sensorManager` variable in cases where the `startTask` method is called multiple times on the same instance.

And is also a bit more robust for future development, since a developer doesn't have to make sure the `sensorManager` is initialized before using it.

Also you can use the generic `context.getSystemService(SensorManager.class)` method so you don't have to do the extra cast like this:

```java
sensorManager = context.getSystemService(SensorManager.class);
```

android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public void startTask() {
+
+        sensorManager = getSensorManager();
```
mvanbeusekom: If you add the initialization of the `sensorManager` into the constructor you can remove this line of code.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
```diff
@@ -0,0 +1,62 @@
...
+class MagnetometerTask extends Task implements SensorEventListener {
+    private final Context mContext;
+
+    private SensorManager sensorManager;
+    Sensor mSensor; // Magnetic sensor returned by sensor manager
+
+    float x; // magnetometer x value
+    float y; // magnetometer y value
+    float z; // magnetometer z value
+    float magnitude; // magnetometer calculated magnitude
```
mvanbeusekom: The above variables are only used in the `onSensorChanged` method so they should be moved into the method.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+import io.flutter.plugin.common.PluginRegistry;
+
+class MagnetometerTask extends Task implements SensorEventListener {
+    private final Context mContext;
+
+    private SensorManager sensorManager;
```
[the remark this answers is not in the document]
audkar: Not really. There is huge movement against using Hungarian Notation. Some [article](https://jakewharton.com/just-say-no-to-hungarian-notation/) by JWharton. And Android Kotlin style explicitly has rule not to use Hungarian notation.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
audkar: getSystemService(class) is added since android 23. I think this code will break at runtime on devices <23
[Docs](https://developer.android.com/reference/android/content/Context#getSystemService(java.lang.Class%3CT%3E))

android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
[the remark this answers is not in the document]
mvanbeusekom: @audkar that is very useful information that I wasn't aware of thanks for that. In that case let's abandon hungarian notation and make sure we do this in all other members as well (meaning we rename `mContext` and `mSensor` to `androidContext` and `sensor`).

android/src/main/java/com/baseflow/flutter/plugin/geolocator/tasks/MagnetometerTask.java
mvanbeusekom: @audkar thanks for that last catch, you are absolutely correct.

martijn00:
We'll move this to it's own plugin.

PR #110, opened 2018-09-16
Use Google Api check plugin to detect play services

martijn00:

Feature

Not able to control location provider

You can now set it

Not to the interface.

- [ ] All projects build


lib/geolocator.dart
```diff
@@ -60,24 +60,24 @@ class Geolocator {
     return _GeolocationStatusConverter.fromPermissionStatus(permissionStatus);
   }
 
+  /// Returns [true] when Google Play Services are installed on device.
+  /// You can override behaviour by setting this property manually.
+  /// When [true] on Android [FusedLocationProviderClient] will be used.
+  /// Falls back to [LocationManager] when set to [false].
+  bool useFusedLocationProvider = GoogleApiAvailability().checkGooglePlayServicesAvailability() == GooglePlayServicesAvailability.success;
```
mvanbeusekom: I think this can go wrong since the `checkGooglePlayServicesAvailability` is an `async` method and it is not awaited here.

Also I choose the name `forceAndroidLocationManager` on purpose. In all programming languages the default value for a boolean is `false` and most developers assume this it the case. So setting a property to `false` feels counterintuitive to most developers (this is also true for switches in a CLI command where the lack of a switch means it is `false`).

lib/geolocator.dart
```diff
@@ -60,24 +60,35 @@ class Geolocator {
...
+  /// On Android devices you can set the parameter [forceAndroidLocationManager]	
+  /// to true to force the plugin to use the [LocationManager] to determine the	
+  /// position instead of the [FusedLocationProviderClient]. On iOS this is ignored.
+  Future<bool> forceAndroidLocationManager([bool forceAndroidLocationManager = false, bool showDialogIfNecessary = false]) async {
+    if(!_checkedGooglePlayServicesAvailability){
+      _forceAndroidLocationManager = await GoogleApiAvailability().checkGooglePlayServicesAvailability(showDialogIfNecessary) == GooglePlayServicesAvailability.success;
+      _checkedGooglePlayServicesAvailability = true;
+    }
+
+    return forceAndroidLocationManager && _forceAndroidLocationManager;
```
martijn00: @mvanbeusekom Will this work? I couldn't find a good sample on how multiple bool checks work.

example/lib/pages/current_location_widget.dart
```diff
@@ -23,9 +23,9 @@ class _LocationState extends State<CurrentLocationWidget> {
     Position position;
     // Platform messages may fail, so we use a try/catch PlatformException.
     try {
+      Geolocator().forceAndroidLocationManager(true);
       position = await Geolocator().getCurrentPosition(
```
mvanbeusekom: This will not work since you are creating a new instance of the `Geolocator` class on both line 26 and line 27 (in Dart you don't have to use the `new` keyword.

You could rewrite it like this :

```dart
Geolocator geolocator = Geolocator();
geolocator.forceAndroidLocationManager(true);
position = await geolocator.getCurrentPosition(
  desiredAccuracy: LocationAccuracy.bestForNavigation);
```

or using the double dot syntax:
```dart
Geolocator geolocator = Geolocator()
    ..forceAndroidLocationManager(true);
position = await geolocator.getCurrentPosition(
  desiredAccuracy: LocationAccuracy.bestForNavigation);
```

lib/geolocator.dart
mvanbeusekom: @martijn00, yes this will work. However I would make this whole method private and add an optional parameter to the constructor of the `Geolocator` class. Also I would leave out the whole `showDialogIfNecessary` and let developers use the `GoogleApiAvailability` plugin themselves if they want to show a dialog.

So as a developer I could do this:

```dart
Geolocator geolocator = Geolocator(forceAndroidLocationManager: true);
geolocator.getCurrentPosition(desiredAccuracy: LocationAccuracy.best);
```

This means you will get code like this in the `Geolocator` class:

```dart
class Geolocator {

Geolocator({this.forceAndroidLocationManager = false});

final bool forceAndroidLocationManager;
bool _checkedGooglePlayServicesAvailability = false;

Future<bool> _shouldForceAndroidLocationManager() async {
    // By doing this check here, we save the App from always checking if Google Play Services
    // are available (which is not necessary if the developer wants to force the use of the LocationManager).
    if (forceAndroidLocationManager) { 
        return true;
    }

    bool googlePlayServicesAvailable = false;
    if (!_checkedGooglePlayServicesAvailability) {
        googlePlayServicesAvailable = await 
            GoogleApiAvailability().checkGooglePlayServicesAvailability(showDialogIfNecessary) == GooglePlayServicesAvailability.success;
        _checkedGooglePlayServicesAvailability = true;
    }
     return !googlePlayServicesAvailable;
}
```

README.md
```diff
@@ -73,12 +73,12 @@ import 'package:geolocator/geolocator.dart';
...
 ```
 
-By default `Geolocator` will use `FusedLocationProviderClient` on Android when Google Play Services are available. It will fall back to `LocationManager` when it is not available.
+By default `Geolocator` will use `FusedLocationProviderClient` on Android when Google Play Services are available. It will fall back to `LocationManager` when it is not available. You can override the behaviour by setting `forceAndroidLocationManager`.
 
 ``` dart
 import 'package:geolocator/geolocator.dart';
 
-bool Geolocator().forceAndroidLocationManager(true);
+bool Geolocator().forceAndroidLocationManager = true;
```
mvanbeusekom: Show complete example here (where you expand on the previous example):

```dart
import 'package:geolocator/geolocator.dart';
 
Geolocator geolocator = Geolocator()..forceAndroidLocationManager = true;
GeolocationStatus geolocationStatus  = await geolocator.checkGeolocationPermissionStatus();
```

PR #152, opened 2018-12-13
Fix Proguard

myleswh:
Proguard fix for Android


mvanbeusekom:
Thanks for contribution!

PR #245, opened 2019-04-14
Add a convenience method to turn a Position into Placemarks.

Lootwig:


lib/geolocator.dart
```diff
@@ -9,10 +9,15 @@ import 'package:meta/meta.dart';
 import 'package:permission_handler/permission_handler.dart';
 
 part 'models/geolocation_enums.dart';
+
```
mvanbeusekom: Please remove the empty line

lib/geolocator.dart
```diff
@@ -9,10 +9,15 @@ import 'package:meta/meta.dart';
...
 part 'models/location_accuracy.dart';
+
```
mvanbeusekom: Please remove the empty line

lib/geolocator.dart
```diff
@@ -9,10 +9,15 @@ import 'package:meta/meta.dart';
...
 part 'models/location_options.dart';
+
```
mvanbeusekom: Please remove the empty line

lib/geolocator.dart
```diff
@@ -9,10 +9,15 @@ import 'package:meta/meta.dart';
...
 part 'models/placemark.dart';
+
```
mvanbeusekom: Please remove the empty line

lib/geolocator.dart
```diff
@@ -9,10 +9,15 @@ import 'package:meta/meta.dart';
...
 part 'models/position.dart';
+
```
mvanbeusekom: Please remove the empty line

lib/geolocator.dart
Lootwig: I was wondering about these. My IDE is set to use dartfmt for formatting and I made no customizations of the style, do you have any clue about fixing this (using IntelliJ)?

mvanbeusekom: That is strange, I always run `flutter format .` in the root of the project (which to my knowledge is an alias of dartfmt) and for me it removes the empty lines. 

Maybe `flutter format .` is more then a simple alias and has a bit different rule set compared to the default dartfmt?

Issue #243, opened 2019-04-12
Geolocator Crashes


marcglasberg:
I'm getting this using **geolocator 3.0.1** on Android.  

I'm not sure, but I believe this error happens when we try to read the location before the last reading has finished. It crashes if you try reading the position 100 times within a loop.

    E/AndroidRuntime( 8677): FATAL EXCEPTION: main
    E/AndroidRuntime( 8677): Process: test.app, PID: 8677
    E/AndroidRuntime( 8677): java.lang.IllegalStateException: Reply already submitted
    E/AndroidRuntime( 8677): 	at io.flutter.view.FlutterNativeView$PlatformMessageHandlerImpl$1.reply(FlutterNativeView.java:197)
    E/AndroidRuntime( 8677): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler$1.success(MethodChannel.java:204)
    E/AndroidRuntime( 8677): 	at com.baseflow.flutter.plugin.geolocator.data.Result.success(Result.java:22)
    E/AndroidRuntime( 8677): 	at com.baseflow.flutter.plugin.geolocator.tasks.LocationUpdatesUsingLocationServicesTask.reportLocationUpdate(LocationUpdatesUsingLocationServicesTask.java:108)
    E/AndroidRuntime( 8677): 	at com.baseflow.flutter.plugin.geolocator.tasks.LocationUpdatesUsingLocationServicesTask.access$000(LocationUpdatesUsingLocationServicesTask.java:19)
    E/AndroidRuntime( 8677): 	at com.baseflow.flutter.plugin.geolocator.tasks.LocationUpdatesUsingLocationServicesTask$1.onLocationResult(LocationUpdatesUsingLocationServicesTask.java:40)
    E/AndroidRuntime( 8677): 	at com.google.android.gms.internal.location.zzau.notifyListener(Unknown Source:4)
    E/AndroidRuntime( 8677): 	at com.google.android.gms.common.api.internal.ListenerHolder.notifyListenerInternal(Unknown Source:17)
    E/AndroidRuntime( 8677): 	at com.google.android.gms.common.api.internal.ListenerHolder$zaa.handleMessage(Unknown Source:5)
    E/AndroidRuntime( 8677): 	at android.os.Handler.dispatchMessage(Handler.java:106)
    E/AndroidRuntime( 8677): 	at android.os.Looper.loop(Looper.java:164)
[...]

PR #251, opened 2019-04-27
Prevent multiple replies on the same method channel result instance

Lootwig:
Got several of these bad bois when my app was running on a bad internet connection:

```
E/AndroidRuntime( 4341): java.lang.IllegalStateException: Reply already submitted
E/AndroidRuntime( 4341): 	at io.flutter.embedding.engine.dart.DartMessenger$Reply.reply(DartMessenger.java:124)
E/AndroidRuntime( 4341): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler$1.success(MethodChannel.java:204)
E/AndroidRuntime( 4341): 	at com.baseflow.flutter.plugin.geolocator.data.Result.success(Result.java:22)
E/AndroidRuntime( 4341): 	at com.baseflow.flutter.plugin.geolocator.tasks.LocationUpdatesUsingLocationServicesTask.reportLocationUpdate(LocationUpdatesUsingLocationServicesTask.java:111)
E/AndroidRuntime( 4341): 	at com.baseflow.flutter.plugin.geolocator.tasks.LocationUpdatesUsingLocationServicesTask.access$000(LocationUpdatesUsingLocationServicesTask.java:21)
E/AndroidRuntime( 4341): 	at com.baseflow.flutter.plugin.geolocator.tasks.LocationUpdatesUsingLocationServicesTask$1.onLocationResult(LocationUpdatesUsingLocationServicesTask.java:42)
E/AndroidRuntime( 4341): 	at com.google.android.gms.internal.location.zzau.notifyListener(Unknown Source:4)
E/AndroidRuntime( 4341): 	at com.google.android.gms.common.api.internal.ListenerHolder.notifyListenerInternal(Unknown Source:17)
E/AndroidRuntime( 4341): 	at com.google.android.gms.common.api.internal.ListenerHolder$zaa.handleMessage(Unknown Source:5)
E/AndroidRuntime( 4341): 	at android.os.Handler.dispatchMessage(Handler.java:106)
E/AndroidRuntime( 4341): 	at android.os.Looper.loop(Looper.java:193)
E/AndroidRuntime( 4341): 	at android.app.ActivityThread.main(ActivityThread.java:6669)
E/AndroidRuntime( 4341): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime( 4341): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
E/AndroidRuntime( 4341): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
I/Process ( 4341): Sending signal. PID: 4341 SIG: 9
Application finished.
```

I did not investigate the plugin's entire architecture regarding the background of `TaskContext`, but it seemed impractical to keep a reference to method channel `Result` objects without invalidating them after use, so that's what this PR is for.


android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/wrapper/MethodResponse.java
```diff
@@ -0,0 +1,38 @@
...
+
+import io.flutter.plugin.common.MethodChannel;
+
+/**
+ * Instances of this class are invalidated once the wrapped {@link MethodChannel.Result}
+ * instance has been used for communication.
+ */
+public class MethodResponse extends ChannelResponse {
+
+    private final MethodChannel.Result result;
```
mvanbeusekom: Rename to match with naming convention:

```suggestion
    private final MethodChannel.Result mResult;
```

android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/wrapper/MethodResponse.java
```diff
@@ -0,0 +1,38 @@
...
+
+    MethodResponse(MethodChannel.Result result) {
+        this.result = result;
+        responseSent = false;
+    }
+
+    @Override
+    public void success(Object object) {
+        synchronized (this) {
+            if (!responseSent) {
```
mvanbeusekom: Instead of keeping track of a boolean we could set `result` to `null` after we communicated the resulting value and then check if the `result` parameter is `null`:

```
            if (!mResult != null) {
                mResult.success(object);
                mResult = null;
            }
```

This has the added benefit of releasing the reference to the result object so the GC can clean it up.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/wrapper/MethodResponse.java
```diff
@@ -0,0 +1,38 @@
...
+                result.success(object);
+                responseSent = true;
+            }
+        }
+    }
+
+    @Override
+    public void error(String code, String message, Object details) {
+        synchronized (this) {
+            if (!responseSent) {
```
mvanbeusekom: Same as with the `success` method

android/src/main/java/com/baseflow/flutter/plugin/geolocator/data/wrapper/MethodResponse.java
Lootwig: Well, GC would still happen when `GeolocatorPlugin.onCompletion()` is called right?

Removed it anyway - I had a discussion once with someone after he had read Clean Code's view on `null` and I guess that was still on my mind here, but I presume it's more about not passing `null` around as replacement for a value.

mvanbeusekom: In general yes GC would happen than but it is always good to clean up references when you don't need them anymore. If someone else forgets to clean up the reference to our wrapper class at least we let go of the `result` handle making sure we don't leak more memory or create faulty situations where the wrong `result` reference is used. 

I loved the Clean Code book by the way :D, and I agree with you that I think they meant not to use `null` as a replacement for a value. In this case we are simply cleaning up a reference we don't need anymore.

mvanbeusekom:
#243

mvanbeusekom:
@Lootwig LGTM, thanks a lot for helping out!

PR #264, opened 2019-05-19
Changed geocoding execution to a separate thread using androids asynctask which handles serial execution of incoming task

pparadox11:
Changed geocoding task to run on a separate thread from main.

Enhancement on Geocoding performance.

Using Geocoding with native views introduces laggy experience, caused by UI thread blocking.
(Example google maps)

Running the Geocoding task on a separated thread, so it won't block the UI thread of Android Platform.

No. All the code that calls the Geocoding API is wrapped on a AsyncTask.

There are no tests submitted. Tested only locally.

- [✔] All projects build
- [✔] Follows style guide lines ([code style guide](https://github.com/BaseflowIT/flutter-geolocator/blob/develop/CONTRIBUTING.md))
- [ ] Relevant documentation was updated. (No needed)

### DIFF
--- a/android/src/main/java/com/baseflow/flutter/plugin/geolocator/utils/MainThreadDispatcher.java
+++ b/android/src/main/java/com/baseflow/flutter/plugin/geolocator/utils/MainThreadDispatcher.java
@@ -33,7 +33,7 @@ public class MainThreadDispatcher {
         });
     }
 
-    public static void dispatchGeocodeResult(@NonNull final ChannelResponse channelResponse, final Object result) {
+    public static void dispatchSuccess(@NonNull final ChannelResponse channelResponse, final Object result) {
         initHandlerIfNull();
         handler.post(new Runnable() {
             @Override
--- a/android/src/main/java/com/baseflow/flutter/plugin/geolocator/utils/MainThreadDispatcher.java
+++ b/android/src/main/java/com/baseflow/flutter/plugin/geolocator/utils/MainThreadDispatcher.java
@@ -1,0 +1,45 @@
+package com.baseflow.flutter.plugin.geolocator.utils;
+
+import android.os.Handler;
+import android.os.Looper;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.baseflow.flutter.plugin.geolocator.data.wrapper.ChannelResponse;
+
+public class MainThreadDispatcher {
+    private static Handler handler = new Handler(Looper.getMainLooper());
+
+    public static void dispatch(Runnable runnable) {
+        initHandlerIfNull();
+        handler.post(runnable);
+    }
+
+    private static void initHandlerIfNull() {
+        if (handler == null) {
+            handler = new Handler(Looper.getMainLooper());
+        }
+    }
+
+    public static void dispatchError(@NonNull final ChannelResponse channelResponse, @NonNull final String channelName,
+                                     @NonNull final String error, @Nullable final Object details) {
+        initHandlerIfNull();
+        handler.post(new Runnable() {
+            @Override
+            public void run() {
+                channelResponse.error(channelName, error, details);
+            }
+        });
+    }
+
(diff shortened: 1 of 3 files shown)


mvanbeusekom:
Hi @pparadox11, thanks for this PR it will be a great improvement to the library. 

I have one problem with the current implementation though (which should be an easy fix, I think). Currently you create a new thread, execute the geocoding calls on this thread and then communicate the results back to Dart also on the new thread.

The Flutter documentation (which required some searching, so I can imagine you missed it) states that all communication on the MethodChannel or EventChannel should occur on the Platform Main Thread (the original thread). 

Quote from [The Engine Argitecture](https://github.com/flutter/flutter/wiki/The-Engine-architecture#threading) (specifically the "Platform Task Runner" paragraph):
> In addition to being the thread on which the embedder interacts with the engine after it is launched, this task runner also executes any pending platform messages. This is handy because accessing most platform APIs is only safe on the platform’s main thread. Plugins don’t have to rethread their calls to the main thread. If plugins manage their own worker threads, it is their responsibility to queue responses back onto the platform thread before they can be submitted back to the engine for processing by Dart code. The rule of always interacting with the engine on the platform thread holds here.

So I think we should make sure that we execute the geocoding logic on a new thread, but marshal the result back to the original thread before communicating back to Dart.

pparadox11:
Hi @mvanbeusekom, thanks a lot for the response.

So we have to execute channel response to the main thread (main looper of the app) as i understood?

mvanbeusekom:
@pparadox11 yes that is correct. This is the thread on which the message originally arrives and one which the current code is executed (so before you introduced the new AsyncTask).

pparadox11:
@mvanbeusekom Ok so i can change it to the same branch and push it again? 
Also i am `Lambda expressions are not supported at language level '7' `, should be updated to support level 8? its not breaking anything if we do update it, as i checked it locally.

mvanbeusekom:
@pparadox11 yes you can simply make the changes on your branch and push them to your Github fork. The changes will be automatically added to this pull-request.

Regarding the lambda expressions, not all Android APIs have support for language level 8 or higher so we prefer not to upgrade to language level 8 since it will reduce the supported Android devices.

pparadox11:
@mvanbeusekom ok i am starting working on a fix.

Ah yes, i am always having on my mind, don't even mind below android 5, it's not worth. But you are right. Thanks.

pparadox11:
@mvanbeusekom Updated. Thanks.

android/src/main/java/com/baseflow/flutter/plugin/geolocator/utils/MainThreadDispatcher.java
```diff
@@ -0,0 +1,45 @@
...
+        initHandlerIfNull();
+        handler.post(new Runnable() {
+            @Override
+            public void run() {
+                channelResponse.error(channelName, error, details);
+            }
+        });
+    }
+
+    public static void dispatchGeocodeResult(@NonNull final ChannelResponse channelResponse, final Object result) {
```
mvanbeusekom: I think a better name here would be `dispatchSuccess` as it could theoretically communicate any kind of success message (not only geocoding results). This would potentially make is easier to reuse the dispatcher.

```suggestion
    public static void dispatchSuccess(@NonNull final ChannelResponse channelResponse, final Object result) {
```

pparadox11: Yeah you are right, its much better, i let the `dispatch` method for general use, but this naming is much better. We can leave `dispatch` method for even more generic use.

pparadox11:
@mvanbeusekom Updated the method calling from the 2 geocode files. Thanks.

pparadox11:
Thanks for your helping too. For the other issue if you remember (Location Services Mode: gps, network, off, high accuracy), i can help with that too if you don't mind.

mvanbeusekom:
Yes that would be great!

pparadox11:
Nice, i will inform you within this week 👍 . Again thanks for your helping. Have a nice Sunday!

Issue #248, opened 2019-04-22
get Current Location Null


BhaveshPrince:
Hello Sir,

while i run app into  android platform getting location null.

Please helm me

xuanjinliang:
Me Too.

My Code: 
`Geolocator geolocator = Geolocator();
Future<Position> getLocation() async {
    Position currentLocation;

    GeolocationStatus permission = await geolocator.checkGeolocationPermissionStatus();
    print('location-->${permission}');

    try {
      currentLocation = await geolocator.getCurrentPosition(
          desiredAccuracy: LocationAccuracy.best);
    } on PlatformException catch (e) {
      print('e-->${e}');
      currentLocation = null;
    }

    print('currentLocation-->${currentLocation}');

    return currentLocation;
  }`

Output:
I/flutter (25071): location-->GeolocationStatus.granted
W/GooglePlayServicesUtil(25071): Google Play Store is missing.

ened:
Try the latest `develop` branch, please.

mvanbeusekom:
@BhaveshPrince, @xuanjinliang, does this issue still occur with the latest version (5.0.0) of the `Geolocator` plugin? 

@xuanjinliang, if you don't have Google Play Store you can try forcing (although this should be done automatically) the plugin to use the legacy `LocationManager` to acquire the location by setting the `forceAndroidLocationManager` property of the `Geolocator` class to `true`:

```dart
try {
  final Geolocator geolocator = Geolocator()
    ..forceAndroidLocationManager = true;
  position = await geolocator.getLastKnownPosition(
    desiredAccuracy: LocationAccuracy.best);
} on PlatformException {
  position = null;
}
```

ronaldG19:
@mvanbeusekom used to work perfectly but got an error earlier.

FalkF:
Having the same problem on Android. it seems like this line in ```getCurrentPosition``` never completes.
```await _methodChannel.invokeMethod('getCurrentPosition',
              Codec.encodeLocationOptions(locationOptions));```

**Comment by FalkF** [comment]
Had the same behaviour with `4.0.3` (nothing happens after the `getCurrentPosition` invocation).
With `5.0.1`I get following platform exception:
```
W/GooglePlayServicesUtil(29584): Google Play Store is missing.
E/MethodChannel#flutter.baseflow.com/geolocator/methods(29584): Failed to handle method call
[...]

Issue #242, opened 2019-04-11
Timeout to stop looking


mulderpf:
## 🏗 Enhancement Proposal

Currently, the request to get a GPS location is "go off and get me the GPS location to this accuracy" and it takes as long as it takes.  On devices with battery saver mode on, it appears that infinity is also a possibility.

In a previous implementation of my app, I could request a GPS location to a specified accuracy with a time limit included.  So it would go off and find the GPS location and return a value as soon as the desired accuracy was met, but if not, it would continue trying to gain more accuracy until the timeout and then simply return whatever it got to at that point.  I would like to see this behaviour again as it worked well for many years.

### Platforms affected (mark all that apply)
- [ ] :iphone: iOS
- [X ] :robot: Android

marcglasberg:
Do you mean you ask for some location and then you may get no answer at all, not even an error?

mulderpf:
If the signal is bad and a fix cannot be obtained at the desired level (e.g. the device only has to rely on wifi or cell phone triangulation), then it will never resolve and just keep trying and trying.  The referenced PR seems to be helping towards this sort of issue.

edihasaj:
+1

Issue #199, opened 2019-02-20
Location is not retrieved by Geolocator


wagnerdelima:
## 💬 Questions and Help
Hello,

I've been using geolocator and ever since the update to version 3.0.0 geolocator can no longer retrieve the location.

I am using the correct dependencies:

> dependencies:
>   geolocator: '^3.0.0'
> 

`targetSdkVersion 28 and compileSdkVersion 28`

Flutter doctor gives me this:
`
[✓] Flutter (Channel stable, v1.0.0, on Mac OS X 10.14.3 18D109, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK 28.0.3)
[✓] iOS toolchain - develop for iOS devices (Xcode 10.1)
[✓] Android Studio (version 3.2)
[✓] IntelliJ IDEA Community Edition (version 2018.2.5)
[✓] Connected device (1 available)

• No issues found!`

Once I call `await Geolocator().getCurrentPosition(desiredAccuracy: LocationAccuracy.high);` the code just doesn't return anything and I have this output in terminal:

`I/DynamiteModule( 4233): Considering local module com.google.android.gms.maps_dynamite:0 and remote module com.google.android.gms.maps_dynamite:221
I/DynamiteModule( 4233): Selected remote version of com.google.android.gms.maps_dynamite, version >= 221
V/DynamiteModule( 4233): Dynamite loader version >= 2, using loadModule2NoCrashUtils
W/System  ( 4233): ClassLoader referenced unknown path: 
W/System  ( 4233): ClassLoader referenced unknown path: /data/user_de/0/com.google.android.gms/app_chimera/m/00000030/n/armeabi-v7a
W/System  ( 4233): ClassLoader referenced unknown path: /data/user_de/0/com.google.android.gms/app_chimera/m/00000030/n/armeabi
I/Google Maps Android API( 4233): Google Play services client version: 12451000
I/Google Maps Android API( 4233): Google Play services package version: 15090018
W/DynamiteModule( 4233): Local module descriptor class for com.google.android.gms.googlecertificates not found.
I/DynamiteModule( 4233): Considering local module com.google.android.gms.googlecertificates:0 and remote module com.google.android.gms.googlecertificates:4
I/DynamiteModule( 4233): Selected remote version of com.google.android.gms.googlecertificates, version >= 4
[...]

Issue #135, opened 2018-11-01
Android Pie (API 28) Geolocator().getCurrentPosition() does not return 


savjolovs:
## 🐛 Bug Report

When I run Geolocator().getCurrentPosition() on Android Emulator with Android Pie (API 28) call is not returning at all.

I run the same code on API 27 location returns fine. I tried to create new emulator, nothing changes.  

### Configuration
Android Emulator with Android Pie (API 28) 

**Version:** 2.1.0

**Platform:** 
- [ ] :iphone: iOS
- [X ] :robot: Android

5lineofcode:
mee too, whats wrong? :(

yostrokon:
Same behaviour here till push some coordinates using extended controls.
Same behaviour with one of two phisycal devices. First with android 5 - works fine, another with 6 - no response till fakegps location engaged.

ChangJoo-Park:
+1 here

HerrNiklasRaab:
Anyone an idea whats the cause? This feature is breaking everything, did anyone tried another plugin?

ibrierley:
Checking the obvious, for those with the problem, do you have a gps signal for it to get a location and have moved ? Just wondering if it's sat there waiting for some new location before it updates ?

solnsumei:
Same issues on oreo except you close and open the app again

ibrierley:
I've been fiddling with an Android Wear watch and had a lot of problems with GPS, but I seem to have it working mostly (it can time out after a long time, but I think that's a different issue).

I'm not sure what exactly has got it working, as I've tried lots of things, but maybe someone can check these off and see if they help. 

0) Make sure your device can get a GPS signal, so outside or feed it locations via the emulator (I feed it a GPX route I downloaded to test).

1) Make sure the screen is on. GPS may stop returning results if it's in minimal mode whatever it's called. I force the screen on when needed with Screen package and Screen.keepOn(true); If the screen is off, it seems to hold and buffer a set of locations until the screen is back on.

2) These are my permissions in AndroidManifest.xml, maybe worth a try
 
   ```<uses-permission android:name="android.permission.ACCESS_LOCATION" />
    <uses-permission android:name="android.permission.INTERNET"/>
    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
    <uses-permission android:name="android.permission.ACCESS_LOCATION_EXTRA_COMMANDS" />
[...]

PR #282, opened 2019-06-13
Current position retrieval timeouts, updated example

ened:

Bugfix & Features

The PR solves a few things, but it started with:
I noticed that my phone would never switch off the GPS antenna when retrieving the location. The root cause was that `geolocator` tried to fetch the position through Androids LocationManager class. My office has no GPS signal, therefore the call would run forever. I attached a timeout to the Dart future (which worked), but the location request would stay because the actual Task was not closed when the Dart Future was cancelled.
As a workaround, I added a timeout option to the getCurrentPosition call and extended the Android Location Task.

To test everything, a few other things were needed:
 - The example needed an additional page for the FusedLocationProvider on Android.
 - The previous implementation of getCurrentPosition would keep the task queue busy by having a `await` call in there, which I tried to avoid.

The PR should probably be split before continuing. Will put it through test.

#### RFC from the maintainer 

Looking back at the PR, I think the introduction of a internal `timeout` field (and the os Handler etc) may not be the best way.
I'd very much prefer to use Darts `.timeout` method on a Future. To use it successfully though, a bit of the task management has to be moved into the Dart side.
For comparison, this PR enables code like this:
``` dart
Geolocator()
  ..forceAndroidLocationManager = false
  ..getCurrentPosition(
    desiredAccuracy: LocationAccuracy.best,
    timeout: Duration(seconds: 5),
  ).then((position) {
    if (mounted) {
      setState(() => _currentPosition = position);
    }
  }).catchError((e) {
    //
  });
```

However, it would be a lot better if Dart could specify the timeout like so:

``` dart
Geolocator()
  ..forceAndroidLocationManager = false
  ..getCurrentPosition(
    desiredAccuracy: LocationAccuracy.best,
  ).timeout(Duration(seconds: 5)).then((position) {
    if (mounted) {
      setState(() => _currentPosition = position);
    }
  }).catchError((e) {
    //
  });
```

Using the below way would be better as the error is not library-defined ("just return null on timeout") and Dart could raise its `TimeoutException` properly. To make that work, the Geolocator API internally would fetch a `TaskID` first, then start a task with that task ID and install a `whenComplete` handler on the returned future to cancel the task.

This comes with another set of problems though - namely what should happen when the Dart side disappears completely (e.g. the UI/ Activity is killed), and the location provider still runs in the background.

Added `timeout` option to `getCurrentPosition` API.

No.

This will need to be tested indoors / outdoors in high GPS and no-GPS situations.

#248 
#242 
#199 
#135 
#123 

- [ ] All projects build

### DIFF
--- a/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java
+++ b/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java
@@ -55,16 +55,35 @@ public class GeolocatorPlugin implements
   @Override
   public void onMethodCall(MethodCall call, Result result) {
     switch (call.method) {
+      case "_createTask":
+        result.success(UUID.randomUUID().toString());
+        break;
+      case "_stopTask":
+        if (call.arguments instanceof String) {
+          UUID taskID = UUID.fromString((String) call.arguments);
+          Task task = mTasks.get(taskID);
+
+          if (task != null) {
+            task.stopTask();
+          }
+          // FWIW, return whether the task has been tracked before.
+          result.success(task != null);
+        } else {
+          result.error("INVALID_TASK_ID", "This is a internal Geolocator error.", null);
+        }
+        break;
       case "getLastKnownPosition": {
+        String taskID = call.argument("taskID");
         Task task = TaskFactory.createLastKnownLocationTask(
-                mRegistrar, result, call.arguments, this);
+            taskID, mRegistrar, result, call.arguments, this);
         mTasks.put(task.getTaskID(), task);
         task.startTask();
         break;
       }
       case "getCurrentPosition": {
+        String taskID = call.argument("taskID");
         Task task = TaskFactory.createCurrentLocationTask(
-                mRegistrar, result, call.arguments, this);
+                taskID, mRegistrar, result, call.arguments, this);
         mTasks.put(task.getTaskID(), task);
         task.startTask();
         break;
--- a/android/src/main/java/com/baseflow/geolocator/data/LocationOptions.java
+++ b/android/src/main/java/com/baseflow/geolocator/data/LocationOptions.java
@@ -1,16 +1,17 @@
 package com.baseflow.geolocator.data;
 
-import java.util.Map;
-
 import static com.baseflow.geolocator.tasks.LocationUpdatesUsingLocationManagerTask.GeolocationAccuracy;
+
+import java.util.Map;
 
 public class LocationOptions {
 
   @GeolocationAccuracy
-  private int accuracy;
-  private long distanceFilter;
-  private boolean forceAndroidLocationManager;
-  private long timeInterval;
+  private final int accuracy;
+  private final long distanceFilter;
+  private final boolean forceAndroidLocationManager;
+  private final long timeInterval;
+  private final long timeout;
 
   public static LocationOptions parseArguments(Object arguments) {
     @SuppressWarnings("unchecked")
@@ -20,16 +21,20 @@ public class LocationOptions {
     final long distanceFilter = (int) map.get("distanceFilter");
     final boolean forceAndroidLocationManager = (boolean) map.get("forceAndroidLocationManager");
     final long timeInterval = (int) map.get("timeInterval");
+    final long timeout = (int) map.get("timeout");
 
-    return new LocationOptions(accuracy, distanceFilter, forceAndroidLocationManager, timeInterval);
+    return new LocationOptions(accuracy, distanceFilter, forceAndroidLocationManager, timeInterval,
+        timeout);
   }
 
 
-  private LocationOptions(@GeolocationAccuracy int accuracy, long distanceFilter, boolean forceAndroidLocationManager, long timeInterval) {
+  private LocationOptions(@GeolocationAccuracy int accuracy, long distanceFilter,
+      boolean forceAndroidLocationManager, long timeInterval, long timeout) {
     this.accuracy = accuracy;
     this.distanceFilter = distanceFilter;
     this.forceAndroidLocationManager = forceAndroidLocationManager;
     this.timeInterval = timeInterval;
+    this.timeout = timeout;
   }
 
   @GeolocationAccuracy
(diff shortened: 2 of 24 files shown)


mvanbeusekom:
@ened, first of all thanks again for your effort. Looking forward on this (big) improvement. 

I fully agree with you that we should try to stick to the Dart timeout implementation. I am not sure yet how we can make sure to tell Android to cancel listening for location updates when the Dart future times out but you seem to have an idea on that so maybe you can point me to some additional reading?

Regarding your question on what to do when the Dart side disappears completely, I would say we can allow users to pass in a cancellation token which that can trigger when the app is disappearing and gives us a little bit of time to cancel the request for a position. How do you feel about that?

ened:
> I fully agree with you that we should try to stick to the Dart timeout implementation. I am not sure yet how we can make sure to tell Android to cancel listening for location updates when the Dart future times out but you seem to have an idea on that so maybe you can point me to some additional reading?

So the getCurrentLocation method (or any other non-stream way) would:
1. Ask the native portion to create a task ID & remember it.
1. Call the native portion and submit the task ID. Native part then puts the task in a hash table of some sort
1. When the task is done, native portion will clear the task
1. When the task times out OR when the task is done, Dart side will call native portion to clear the task using [`whenComplete`](https://api.dartlang.org/stable/2.1.0/dart-async/Future/whenComplete.html).
    - If the task is already done (& therefore cleared), then nothing need to be done
    - If the task is still running, it will now be cancelled
 
> Regarding your question on what to do when the Dart side disappears completely, I would say we can allow users to pass in a cancellation token which that can trigger when the app is disappearing and gives us a little bit of time to cancel the request for a position. How do you feel about that?

I'm not sure about cancellation tokens and I prefer to not add external APIs or management (*users to pass token* sounds like it). Anything developers have to manage is something we'd need to test, maintain, document.

Android we can receive install (`onViewDestroy`)[https://api.flutter.dev/javadoc/io/flutter/plugin/common/PluginRegistry.ViewDestroyListener.html] for the moment) listener which should be enough for this use case.

Will continue thinking - and may submit further PR to prepare for the changes.

ened:
@mvanbeusekom current implementation pushed, happy to hear your comments.

ened:
Another thing we may want to show in the example project is how to cancel a location request. It should done by [cancelling a stream](https://dart.academy/how_cancel_future/).

mvanbeusekom:
@ened I have reviewed the code and I really like the way you approached it. Up to here it looks really good and I don't have any remarks so far. The PR is still annotated with [WIP] so I haven't approved it yet.

Regarding your remark on extending the example project with an example on how to cancel a location request. I think that would be a good idea, but maybe in a separate PR if you don't mind. It would make it easier for me to review and keep things apart.

ened:
hi @mvanbeusekom I have removed the WIP label and have shipped our production App with these changes for a few weeks now. I have confidence the code works, but at some point would like this project to have proper e2e testing as seen in flutter/plugins.

Perhaps a next version with this could be labelled as release candidate (or major version bump) to avoid breaking apps.

ened:
As a side note, I think the project could benefit from some sort of CI / automatic build, test and code formatting checks. :-)

mvanbeusekom:
> As a side note, I think the project could benefit from some sort of CI / automatic build, test and code formatting checks. :-)

We are using Bitrise to automatically build, test and analyze the code. Unfortunately the application on Bitrise is created as a "private" build instead of "public" which means you need to be invited in out team to be able to see the build results. We are in the process of correcting this, however Bitrise doesn't support simply updating the App from "private" to "public" so we basically need to delete the current configuration and create a new App as "public" and add the same config.

mvanbeusekom:
@ened I have finished the conversion from a "Private" Bitrise project into a "Public" Bitrise project. This means you can now also see the results of the automated builds. 

Each PR will be checked automatically and links to the logs will be added to the PR. Because this PR is created before we added the Bitrise integration is doesn't have any checks (I will see if I can still add them).

ened:
@mvanbeusekom saw it - but the error is not related to this PR. Will open a second PR to fix the analysis issues. Thx!

Do you run flutter test / flutter driver on bitrise as well?

ened:
@mvanbeusekom turns out the analysis errors did come from my PR; now resolved and ready for review once again.

audkar:
I have some concerns regarding these changes. (And geolocator's `Task` complexity overall). And they makes things even more complicated.

So this PR does two things: 
1. Tries to solve problem that `getCurrentPosition()` doesn't remove locationListener when `Future` is canceled.
2. Adds new feature to have timeout parameter.

## Problem nr1:
Because of Flutter `Feature` can't be canceled and Flutter IPC with native code doesn't allow serial calls it is quite hard to know when Androids locationListener should be removed.

I think correct way to solve this would be to drop current `getCurrentPosition()` implementation (which is based on `MethodChannel`) and reuse technique used for `getPositionStream()`.  `getPositionStream()` is implemented with `EventChannel`. So native plugin gets notified when no more subscribers left and can remove locationUpdateListener from location manager.

Under the hood API's `getCurrentPosition()` method can be like: 
```
CancelableOperation<Location> getCurrentPosition(args) {
  final streamController = StreamController<Location>();
  return CancelableOperation.fromFuture(Future(() {
    final locations = getLocationStream(args).take(1);
    locations.pipe(streamController.sink);
    return streamController.stream.first;
  }), onCancel: () {
    streamController.close();
  });
}
```
So all Geolocator plugin logic would be: 
- have single location BroadcastStream
- connect via StreamChannel
- on Android side add location listener when streamChannel has at least one subscriber
- remove location listener when streamChannel has no subscribers

## Problem nr2:
I see no point on implementing timeout feature in such complex way when it is possible to do smth like:
```
getLocationStream(args).take(1).timeout(duration);
```

ened:
Hi @audkar thank you that's a great review and a great suggestion to improve it.

Will update the PR asap.

arcturus:
Any plans to have this merged?

mvanbeusekom:
I have added time out support based on streams in version 6.0.0 (which is currently available as release candidate)

PR #350, opened 2019-11-08
Fix plugin response key for position and update example.

slightfoot:


slightfoot:
@mvanbeusekom this would not have happened of their was a mock plugin handler and tests 🤔😇

mvanbeusekom:
@slightfoot luckily you added one so we prevent it from happening again 🤔😇

Issue #366, opened 2019-12-25
Bounding box parameter for placemarkFromAddress() method


sanisloandras:
## 🚀 Feature Requests

Hello!

What I want to achieve is to geocode addresses into geographic coordinates ONLY within a specific city.
I'm not sure what's the base idea of this library and what does it use internally, but, for example, Android's Geocoder library supports a bounding box for the search results. Read more about this [here](https://developer.android.com/reference/android/location/Geocoder.html#getFromLocationName(java.lang.String,%20int,%20double,%20double,%20double,%20double)).

### Platforms affected (mark all that apply)
- [x] :iphone: iOS
- [x] :robot: Android

sanisloandras:
Submitted a PR: https://github.com/Baseflow/flutter-geolocator/pull/367

mvanbeusekom:
Since version 6.0.0 the geocoding features are no longer part of the geolocator plugin and moved to their own plugin which can be found here.

PR #367, opened 2019-12-26
Feature/bounding box

sanisloandras:
Solves the following issue: https://github.com/Baseflow/flutter-geolocator/issues/366

There are no parameters for bounding box of placemarkFromAddress method.

There are additional parameters of the placemarkFromAddress method for bounding box.

no

- [x] All projects build


mvanbeusekom:
@sanisloandras thank you very much for you contribution. I like the idea and appreciate your pull request.

For Android this looks fine and now I have been looking if we could also support this on iOS. Researching this a bit I came to the conclusion that on iOS you could do something similar only instead it uses a circular based region (center coordinate + radius).

I was wondering if you would mind if I pick up on this idea and make some modifications so that we can use both features. I would probably change the signature a little bit so it is clear which platform and which bounding limitation we are using (e.g. the bounding box on Android and the Circular region on iOS)?

Issue #390, opened 2020-02-10
Location in Background Mode 


thomasnilsson:
## 🚀 Feature Requests

At least on iOS, the app will stop tracking the location after about 30 seconds, after the application goes into the background.

### Contextualize the feature
Continually track the location

### Describe the feature
Add this line after line 38 in `LocationTask.m`
```
 _locationManager.allowsBackgroundLocationUpdates = YES;
```
Not sure what the status is on Android.

### Platforms affected (mark all that apply)
- [ x ] :iphone: iOS
- [ ] :robot: Android

hlongngo:
I am having the same issue.

estevez-dev:
The app can't be launched in the background forever. To save user's battery and memory iOS and Android will close your application been some time on the background. You can try to use [workmanager](https://pub.dev/packages/workmanager) or [foreground_service](https://pub.dev/packages/foreground_service) for this. Nothing to do in geolocator plugin itself.

estevez-dev:
Also see some comments in #389

RomanJos:
Would it be possible to at least have update if we lock the phone ?

SetNameHere:
Does this pull request only address the issue for IOS?
On android I get the same behaviour.
After some time the app will be terminated, preventing the receiption of any stream updates.

thomasnilsson:
> Does this pull request only address the issue for IOS?
> On android I get the same behaviour.
> After some time the app will be terminated, preventing the receiption of any stream updates.

Google has recently changed the behaviour of the location API, read this: https://developer.android.com/about/versions/oreo/background-location-limits

SetNameHere:
So if I get it right, we need in addition to your IOS fix a foreground service, handling the stream functionality on android side, correct?

thomasnilsson:
Yep. This plugin (https://pub.dev/packages/foreground_service) worked for me, although running a foreground service drains the battery quite a lot in my experience, and I had some compilations issues when compiling for iOS, when using the package.

SetNameHere:
I have switched for the moment to the background_locator plugin (https://pub.dev/packages/background_locator#-readme-tab-).
Currently it works realy good.

PR #412, opened 2020-03-25
Location updates in Background Mode

thomasnilsson:
Now location updates can be streamed while app is in the background, i.e. when the user is using another app which is in the foreground. 

Currently, the location stream updates will stop after about 30 seconds of the app going into the background. I don't see a use case where this is a desired streaming behavior, but I could be wrong.

The user will need to create the following entries in the plist:
``` xml
<key>NSLocationWhenInUseUsageDescription</key>
<string>This app needs access to location when open.</string>
<key>NSLocationAlwaysUsageDescription</key>
<string>This app needs access to location when in the background.</string>
<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
<string>This app needs access to location when open and in the background.</string>
```

In addition, the user will need to add the `Background Modes` capability to their XCode project (Project > Signing and Capabilties > "+ Capability" button) and select `Location Updates`.

All in all, the user can no longer choose to only allow foreground updates.

If the permissions described in the previous section are not given, the app will crash.

https://github.com/Baseflow/flutter-geolocator/issues/390

- [x ] All projects build
- [x ] Follows style guide lines ([code style guide](https://github.com/BaseflowIT/flutter-geolocator/blob/develop/CONTRIBUTING.md))
- [ x] Relevant documentation was updated
- [x ] Rebased onto current develop

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,22 +165,18 @@ On Android you'll need to add either the `ACCESS_COARSE_LOCATION` or the `ACCESS
 
 ### iOS
 
-On iOS you'll need to add the `NSLocationWhenInUseUsageDescription` to your Info.plist file (located under ios/Runner) in order to access the device's location. Simply open your Info.plist file and add the following:
+On iOS you'll need to add the following entries to your Info.plist file (located under ios/Runner) in order to access the device's location. Simply open your Info.plist file and add the following:
 
 ``` xml
 <key>NSLocationWhenInUseUsageDescription</key>
 <string>This app needs access to location when open.</string>
-```
-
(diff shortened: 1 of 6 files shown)


ramsestom:
Background location should not be on by default. I should be an option of the getPositionStream function especially since the app would crash if the background mode permission is not set.
(you can take example to this lib https://github.com/loup-v/geolocation)
Bytheway, to enable background mode on iOS you have to justify this usage to Apple and Apple can reject your app if they consider it not to be fully justified to improve user experience. So most developpers won't ask for this permission unless they really need it.

mvanbeusekom:
@ramsestom thank you for thinking a long, it does makes sense to make this configurable (and disable it by default). I have created a separate issue #423 to keep track of progress on this suggestion.

PR #463, opened 2020-07-02
Added checkPermissions to Android

renefloor:
Feature

No platform code for Android

Check Permission logic on Android Platform

No

I tested the example app.
It said it had no permissions. After turning them on it set 'last known location: null'
I needed to add a catch for MissingPluginException in the example and the `int.toLocationPermission()` needed an extra value. 

- [X] All projects build
- [X] Follows style guide lines ([code style guide](https://github.com/BaseflowIT/flutter-geolocator/blob/develop/CONTRIBUTING.md))
- [X] Relevant documentation was updated
- [X] Rebased onto current develop

### DIFF
--- a/geolocator/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
+++ b/geolocator/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
@@ -1,0 +1,119 @@
+package com.baseflow.geolocator.permission;
+
+import android.Manifest;
+import android.app.Activity;
+import android.content.Context;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.os.Build;
+
+import androidx.annotation.Nullable;
+import androidx.core.content.ContextCompat;
+
+import com.baseflow.geolocator.errors.ErrorCallback;
+import com.baseflow.geolocator.errors.PermissionUndefinedException;
+
+import java.util.Map;
+
+import io.flutter.plugin.common.PluginRegistry;
+
+public class PermissionManager {
+    @FunctionalInterface
+    public interface ActivityRegistry {
+        void addListener(PluginRegistry.ActivityResultListener handler);
+    }
+
+    @FunctionalInterface
+    public interface PermissionRegistry {
+        void addListener(PluginRegistry.RequestPermissionsResultListener handler);
+    }
+
+    @FunctionalInterface
+    public interface RequestPermissionsSuccessCallback {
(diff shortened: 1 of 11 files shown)


geolocator/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @FunctionalInterface
+    public interface RequestPermissionsSuccessCallback {
+        void onSuccess(Map<Integer, Integer> results);
+    }
+
+    @FunctionalInterface
+    public interface CheckPermissionsSuccessCallback {
+        void onSuccess(int permissionStatus);
+    }
```
mvanbeusekom, line 39: I think you could make one interface here since we will not return multiple permission states. We will always just return the permission status for the location permission. So maybe do something like:

```suggestion
    @FunctionalInterface
    public interface PermissionsSuccessCallback {
        void onSuccess(int permissionStatus);
    }
```

renefloor: I guess we should return the enum than as well instead of an int?
```suggestion
    @FunctionalInterface
    public interface PermissionsSuccessCallback {
        void onSuccess(LocationPermission permissionStatus);
    }
```

mvanbeusekom: yes I think that would be good. That would mean that we need to do the conversion to int in the `MethodHandlerImpl` but I think that is correct since that is the class that is responsible for communication over the method channel.

geolocator/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @FunctionalInterface
+    public interface PermissionsSuccessCallback {
+        void onSuccess(LocationPermission permissionStatus);
+    }
+
+    @FunctionalInterface
+    public interface ShouldShowRequestPermissionRationaleSuccessCallback {
+        void onSuccess(boolean shouldShowRequestPermissionRationale);
+    }
```
mvanbeusekom, line 39: I think this one is not necessary, the `shouldShowRequestPermissionRationale()` Android method is a synchronous method which will return a value immediately so no need to have a separate callback. In other words, we can treat this similar as the `checkPermission` method.

PR #490, opened 2020-08-26
Update functions documentation

hugocbpassos:
Some functions from `geolocator_platform_interface.dart` are exposed through `geolocator.dart`, so they share the documentation. It seems like `getLastKnownPosition()`'s documentation was updated in `geolocator_platform_interface.dart`, but kept outdated in `geolocator.dart`, so I sorted that out.

Furthermore, `getPositionStream()`'s documentation doesn't contemplate geolocator's 6.0.0 refactor, since it instantiates `Geolocator` class, which no longer exists. I also updated that one.

### DIFF
--- a/geolocator/lib/geolocator.dart
+++ b/geolocator/lib/geolocator.dart
@@ -28,8 +28,10 @@ Future<bool> isLocationServiceEnabled() =>
 
 /// Returns the last known position stored on the users device.
 ///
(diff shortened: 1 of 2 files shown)


mvanbeusekom (approved):
That is awesome, thank you very much for taking the time to correct it!

PR #528, opened 2020-09-23
Fused location provider

danielroek:

- [ ] All projects build

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ Setting up your development environment
 Running the example project
 ---------------------------
 
- * Change into the example directory: `cd example`
+ * Change into the example directory: `cd geolocator/example`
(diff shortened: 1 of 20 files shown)


danielroek:
To reproduce and test this issue, I made a fresh API 28 device. Added the following method to the initState and a button. 
```  
_test() async {
    debugPrint((await isLocationServiceEnabled()).toString());
  }
```
When starting a fresh API 28 device, location is always on. You can find this by opening the notification slider, and adding the location button to the menu.  So when the app opens, location services are on. Which is why isLocationServiceEnabled returns true. 

If No Thanks is clicked in the android notification, Location services is still on, it just doesn't use Wifi etc. to refine the location. 
If I create a fresh API 28 device, open the notification slider, disable Location services, open the app, my test code returns false.

mvanbeusekom (approved):
LGTM!

PR #542, opened 2020-10-08
Update README.md

rogiervandenberg:
The way ACCESS_COARSE_LOCATION works on Android is slightly different than described in the Readme. 

- To let your app GPS himself, it needs to call `getCurrentPosition();` and therefore needs the permission `ACCESS_FINE_LOCATION`
- To let your app ask the phone if he knows where he is, you can use `getLastKnownPosition();` while having permission `ACCESS_COARSE_LOCATION`.

_Having set just ACCESS_COARSE_LOCATION and trying to get the current position based on GPS will hang the process._

**Therefore I have changed:**

* The NOTE below the permissions, to explain that ACCESS_COARSE_LOCATION only works with `getLastKnownPosition`
* A note below the usage of the `getCurrentPosition` method to explain that `ACCESS_FINE_LOCATION` is necessary for Android. This was driving me nuts ;)

It is a docs update

Essential stuff was missing, which lead to mis-using the plugin and therefore not working.

More explicit setup. 

Nope..

Using `<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />` with 
`Position position = await getCurrentPosition();` does not work and hangs. The phone then tries to GPS, but cannot while having just COARSE permissions. Even with `LocationAccuracy.lowest`

I had a 1-on-1 chat with Maurits

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Starting from Android 10 you need to add the `ACCESS_BACKGROUND_LOCATION` permis
 <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
 ```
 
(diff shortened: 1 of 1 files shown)


mvanbeusekom (approved):
LGTM! Thanks

Issue #547, opened 2020-10-13
Example on pub.dev is misleading. Needs to display code having the proper Geolocator plugin import


Sakshi725744:
### Pitch

The given example on pub.dev is not portraying the example carrying the import for the plugin, i.e. import 'package:geolocator/geolocator.dart'; is nowhere added making the documentation having uncertainties. I myself faced this confusion when i came across the example page and hence "main.dart" folder needs modifications.

### Platforms affected (mark all that apply)
- [x] :iphone: iOS
- [x] :robot: Android

Sakshi725744:
I would like to volunteer for this issue

mvanbeusekom:
Looking forward to your PR!

icelab-uki:
Totally agree the example not include the presented library.
And it's full of unknown error.

I tried to run
Position newPosition = await Geolocator.getCurrentPosition();

But it keep saying
MissingPluginException(No implementation found for method isLocationServiceEnabled on channel

Really frustated

pp2015jp:
Totally agree the example not include the presented library.
And it's full of unknown error.

I tried to run
Position newPosition = await Geolocator.getCurrentPosition();

But it keep saying
MissingPluginException(No implementation found for method isLocationServiceEnabled on channel

Really frustated

mvanbeusekom:
I have been looking into this and the problem is that pub.dev only shows the code in the `example/lib/main.dart` file.

Our accompanying example app however has a much more detailed example with multiple code files showing different approaches (using `getLastKnownPosition` and `getPositionStream` for example).

So if you are looking for examples please open the complete example app and look a bit further then simple the "Example" tab on pub.dev.

Regarding the other error you are receiving, if you are on Android make sure you update your gradle version. If you want more help I'd suggest opening a separate issue.

pp2015jp:
Thank you very much for your reply.
But the example is not runnable.
And it is difficult for me to follow because it uses
"FutureBuilder(
      future: Geolocator.getLastKnownPosition(),
..."
But I did not use FutureBuilder in my work.

I only need to know how to get location as a variable, which is suppose to be done easily by
"Position position = await Geolocator.getCurrentPosition(desiredAccuracy: LocationAccuracy.high);"
as written on https://pub.dev/packages/geolocator

But it's not true. I faced many errors and it already took me almost two whole days, just to get simple location data.

Here are examples of error

1.
[...]

PR #586, opened 2020-11-06
improved example app to minimize code that is not relevant

Sempakonka:

An improved example app.

An example app wich is less clear, and can cause confusion.

This does not introduce new behavior.

No.

When running the example app, you will be able to see the changes.

[Fixes Baseflow/flutter-geolocator#547](https://github.com/Baseflow/flutter-geolocator/issues/547)

- [x] All projects build


geolocator/example/lib/main.dart
```diff
@@ -1,163 +1,155 @@
...
+            return Card(
+              child: ListTile(
+                tileColor: themeMaterialColor,
+                title: Text(posList[index].toString(),
+                    style: TextStyle(color: Colors.white)),
               ),
             );
-          }
-        }()),
+          }),
```
mvanbeusekom, line 60: This code is a bit confusing. Basically you are altering the actual amount of elements in the array to be able to put an initial item containing the permission status in the `ListView`. This is very confusing and makes it prone to mistakes (since you need to keep track of the index and change it a few times).

It would be a lot easier if we change the type of items in the list to match with what we would like to display. If we simply change the list to hold a `String` we can update this code to:

```dart
body: ListView.builder(
  itemCount: posList.length,
  itemBuilder: (context, index) {
    return ListTile(
      tileColor: themeMaterialColor,
      title: Text(posList[index],
        style: TextStyle(color: Colors.white)),
    );
  },
);
``` 

If you want to use a different display element for each type of item in the list we could create our own class:

```dart
enum _PositionItemType {
  permission,
  position,
}

class _PositionItem {
  _PositionItem(this.type, this.displayValue);

  final PositionItemType type;
  final String displayValue;
}
```

Update the `posList` to hold a list of `PositionItem` and update the `ListView.builder` like so:

```dart
body: ListView.builder(
  itemCount: posList.length,
  itemBuilder: (context, index) {
    final positionItem = posList[index];
   
    if (positionItem.type == PositionItemType.permission) {
      return ListTile(
        textAlign: TextAlign.center
        title: Text(positionItem.displayValue,
          style: TextStyle(
            color: Colors.white,
            fontWeight: FontWeight.bold,
          )),
      );
    } else {
      return Card(
        child: ListTile(
          tileColor: themeMaterialColor,
          title: Text(positionItem.displayValue,
            style: TextStyle(color: Colors.white),
          ),
        ),
      );      
    }
  },
);
```

Of course this means you will also need to change the way how items are added to the list, but this should be straight forward.

geolocator/example/lib/main.dart
```diff
@@ -1,163 +1,155 @@
...
+      _positionStreamSubscription.pause();
+    }
+
+    setState(() {
+      if (_positionStreamSubscription.isPaused) {
+        _positionStreamSubscription.resume();
+      } else {
+        _positionStreamSubscription.pause();
+      }
+    });
```
mvanbeusekom, line 153: It would be good to add an dispose method where you cancel the _positionStreamSubscription if it isn't `null`:

```dart
@override
void dispose() {
  if (_positionStreamSubscription != null) {
    _positionStreamSubscription.cancel();
    _positionStreamSubscritpion = null;
  }
}

File: geolocator/example/pubspec.yaml [hunk 3]
```diff
@@ -42,4 +46,10 @@ flutter:
...
     - res/images/baseflow_logo_def_light-02.png
-    - res/images/poweredByBaseflowLogoLight@3x.png
\ No newline at end of file
+    - res/images/poweredByBaseflowLogoLight@3x.png
+    - packages/baseflow_plugin_template/logo.png
+    - packages/baseflow_plugin_template/poweredByBaseflow.png
+
+
+
+
```
mvanbeusekom, line 55: nit: it would be nice to remove these empty lines:

```suggestion
```

mvanbeusekom (changes requested):
I have a few remarks regarding the code structure which can be simplified a bit. Since this is an example app we should strive to keep the code as clean and easy to understand as possible (actually we should always strive to do this ;))

geolocator/example/lib/main.dart
```diff
@@ -1,163 +1,192 @@
...
 import 'package:flutter/material.dart';
-import 'package:flutter/widgets.dart';
+import 'package:geolocator/geolocator.dart';
 
-import 'settings_menu.dart';
-import 'template/globals.dart' as globals;
+enum _PositionItemType {
+  permission,
+  position,
+}
```
mvanbeusekom, line 10: nit: it would be better to move this enum close to the `_PositionItem` class (at the bottom of the file). 

The enum is just an implementation detail and not directly related to how the `Geolocator` functions. Placing it at the top of the code file it just distracts from the real example code which is in the `GeolocatorWidget`.

mvanbeusekom (approved):
One small suggestion on repositioning the `_PositionItemType` enum.

PR #591, opened 2020-11-12
Use 1 for Android activity request code. Closes #590

### LINKED ISSUE #590: IllegalArgumentException: Can only use lower 16 bits for requestCode
## 🐛 Bug Report

In some circumstances, it is possible to see Exception like this:
```
Fatal Exception: java.lang.IllegalArgumentException: Can only use lower 16 bits for requestCode
       at androidx.fragment.app.d.F(:10)
       at androidx.fragment.app.d.startIntentSenderForResult(:7)
       at com.google.android.gms.common.api.Status.n(:19)
       at com.google.android.gms.common.api.i.b(:2)
       at c.b.a.m.j.o(:23)   
       at c.b.a.m.b.d(:8)
       at c.d.a.b.i.u.run(:27)
       at android.os.Handler.handleCallback(Handler.java:873)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:201)
       at android.app.ActivityThread.main(ActivityThread.java:6810)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:547)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:873)
```
c.b.a.m.j.o(:23)   looks like points to FusedLocationClient:160 `rae.startResolutionForResult(activity, activityRequestCode);`

If this is the case, we should not initialize `activityRequestCode` as `this.hashCode();` because, obviously, it uses more than 16bits.

**Version:** 6.1.5

**Platform:** 
- [x] :robot: Android

corcoran:
Bug fix

Throws exception on Android

Removes possibility of the request code not being a lower 16 bit int

No

https://stackoverflow.com/questions/33331073/android-what-to-choose-for-requestcode-values

- [x ] All projects build
- [x ] Follows style guide lines ([code style guide](https://github.com/BaseflowIT/flutter-geolocator/blob/develop/CONTRIBUTING.md))
- [x ] Relevant documentation was updated
- [x ] Rebased onto current develop

### DIFF
--- a/geolocator/android/src/main/java/com/baseflow/geolocator/location/FusedLocationClient.java
+++ b/geolocator/android/src/main/java/com/baseflow/geolocator/location/FusedLocationClient.java
@@ -30,7 +30,7 @@ class FusedLocationClient implements LocationClient {
     this.context = context;
     this.fusedLocationProviderClient = LocationServices.getFusedLocationProviderClient(context);
     this.locationOptions = locationOptions;
-    this.activityRequestCode = this.hashCode();
+    this.activityRequestCode = 1;
 
     locationCallback =
         new LocationCallback() {


mvanbeusekom (changes requested):
Unfortunately this solution will cause a series of other problems:

1. First of all, the geolocator can (potentially) run multiple requests to enable location services at the same. Imagine someone if running the `getCurrentPosition` and `getPositionStream` at the same time. This means both will trigger a request to enable location services resulting in an activity request. To distinguish the requests from each other, we need to have there request code to be different from each other. 

2. The second problem is that other plugins are also already using request code 1. Which means our handler (or their handler) will also be triggered on events. Now we will never really know if someone else is already using a certain number, but using 1 is the biggest change we have a match.

corcoran:
Good catches! I see the other PR is up, so I'll close this

mvanbeusekom:
Thank you very much though pointing me in the right direction! If you are ok with it I would like to add you as co-author of the other PR

corcoran:
Sounds good 👍

PR #630, opened 2020-12-20
[Android]: Improve retrieving the current location(new Android API)

XinyueZ:

- The latest improvements to the Fused Location Provider API (FLP).
   - **Prerequisites**: Android API Level >v27
- compileSdkVersion -> 30
- Gradle update2date

In order to get the current position, it has to subscribe to ongoing location changes.

Developers wanted an easier way to retrieve the current location. With the new `getCurrentLocation`() API (**Android API Level >v27**)  developers can get the current location in a single request, rather than having to subscribe to ongoing location changes. By allowing developers to request location only when needed (and automatically timing out and closing open location requests), this new API also improves battery life.

A sample of Kotlin https://github.com/android/location-samples/tree/main/CurrentLocationKotlin
 
- [x] All projects build

### DIFF
--- a/geolocator/android/build.gradle
+++ b/geolocator/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.0'
+        classpath 'com.android.tools.build:gradle:3.5.4'
     }
 }
 
@@ -27,7 +27,7 @@ project.getTasks().withType(JavaCompile){
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 30
 
     defaultConfig {
         minSdkVersion 16
@@ -42,5 +42,5 @@ android {
 }
 
 dependencies {
-    implementation 'com.google.android.gms:play-services-location:17.0.0'
+    implementation 'com.google.android.gms:play-services-location:17.1.0'
 }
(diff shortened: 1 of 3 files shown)


mvanbeusekom:
Hi @XinyueZ, 

Thank you for your interest in de geolocator plugin and the effort to make improvements. However the changes you are suggesting don't really do anything. You have declared (and implemented) a new private method `getCurrentLocation` which is never called. 

Also the `getCurrentLocation` method in de `FusedLocationProviderClient` is just a convenience method which underneath will simply request location updates and stop them when the first updated arrived. This is exactly the same the geolocator is doing so I don't really see the need for this update. In general it will make the code more complex, since now we need to maintain the logic for subscribing to the location updates and we need to maintain the added code to request the current location. The geolocator offers a developer friendly `getCurrentPosition` method on the Dart side which will handle the subscribing, unsubscribing and monitoring the timeout for you so you don't have to worry about all that. 

Another problem is that we still want to support the older API levels (keep in sync with Flutter, which currently supports minimum API level 18). Implementing the `getCurrentlLocation` method will limit us to API level 27, which really is a no go. So this result we need to maintain again two ways of supporting users acquiring the current location, and since the `getCurrentLocation` implementation in the FusedLocationProviderClient does exactly the same we do now, I don't see the need to increase the complexity of the code base.

Regarding the update of Gradle. In the past we have experienced a lot of issues due to the fact that we used a different version of Gradle compared to the Flutter stable channel. So for now I wish to adhere to the same version Flutter is currently using on the stable channel.

If there would be any additional benefits to splitting up the logic that I am not aware of please let me know. I would be happy to reconsider if there is any added value. For now I will close this PR, but with the right motivation I am happy to re-open it.

PR #643, opened 2021-01-15
Document the need to set EnableBackgroundLocationUpdates to enable background location tracking on iOS

willbryant:

Docs update

Following the existing instructions in README was not enough to enable background location updates on iOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,13 @@ On iOS you'll need to add the following entries to your Info.plist file (located
 <string>This app needs access to location when in the background.</string>
 ```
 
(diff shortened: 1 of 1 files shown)


mvanbeusekom (commented):
In general you are right, the iOS code of the geolocator indeed checks for this boolean to be present in the `Info.plist`. However I now realise that this is completely not necessary since adding the Background Modes capability through Xcode already adds the entries below to the `Info.plist`:

```
	<key>UIBackgroundModes</key>
	<array>
		<string>location</string>
	</array>
```

Looking back at this it would make much more sense to update the iOS code of the Geolocator plugin to scan for this key instead of defining our own new key.

So this makes me a bit reluctant to accept your changes. Even though you are 100% correct that this is what is needed, the better solution would be to update the iOS code [here](https://github.com/Baseflow/flutter-geolocator/blob/42da68b330cf857cbfbf106c489d72b37779c1d2/geolocator/ios/Classes/Handlers/GeolocationHandler.m#L87).

willbryant:
Makes sense.

PR #738, opened 2021-05-27
Callback locationservice platform interface

florissmit1:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)\
Introducing a feature which returns a callback when the user manually disables / enables location services on their device. The user can listen to the location services stream which will return a boolean based on the location services status.

Currently it is not possible to do such a thing. 

Since it's a new feature, a breaking change is introduced.

Use the example app.


geolocator_platform_interface/lib/src/implementations/method_channel_geolocator.dart
```diff
@@ -21,13 +21,19 @@ class MethodChannelGeolocator extends GeolocatorPlatform {
...
+      EventChannel('flutter.baseflow.com/geolocator_service_updates');
+
   /// On Android devices you can set [forceAndroidLocationManager]
   /// to true to force the plugin to use the [LocationManager] to determine the
   /// position instead of the [FusedLocationProviderClient]. On iOS this is
   /// ignored.
   bool forceAndroidLocationManager = false;
 
   Stream<Position>? _positionStream;
+  Stream<bool>? _serviceStatusStream;
```
mvanbeusekom, line 36: It would be better if we could return an enum value explaining the state instead of a boolean. So let's define a `ServiceStatus` enum that looks something like this:

```dart
/// Describes the current state of the location service on the native platform.
enum ServiceStatus {
  /// Indicates that the location service on the native platform is enabled.
  enabled,
  /// Indicates that the location service on the native platform is enabled.
  disabled,
}
```

And make sure this is wat we return on the stream:
```suggestion
  Stream<ServiceStatus>? _serviceStatusStream;
```

This makes the code self-explanatory and makes it easier for developers to understand what a certain result means.

mvanbeusekom (changes requested):
Looks great, just one small remark to improve the readability of the code.

mvanbeusekom (changes requested):
@florissmit1 good to see you updated the stream to return a `ServiceStatus` enum. You forgot however to also update the unit-tests. They are still expecting a `bool` instead of the new `ServiceStatus` enum. 

Once you have fixed this I think the PR is good to go.

mvanbeusekom (approved):
LGTM

PR #742, opened 2021-06-03
Callback locationservice geolocator

florissmit1:
Feature, docs updated and example app edited

At the moment it's not possible to track if the location service of the device is enabled/disabled 

By starting the serviceStatusStream, an event will be fired when the user turns on/off the devices location service in the settings or in the notification bar

No

Use the (new) example application

### DIFF
--- a/geolocator/android/src/main/java/com/baseflow/geolocator/LocationServiceHandlerImpl.java
+++ b/geolocator/android/src/main/java/com/baseflow/geolocator/LocationServiceHandlerImpl.java
@@ -1,0 +1,61 @@
+package com.baseflow.geolocator;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.location.LocationManager;
+import android.util.Log;
+
+import androidx.annotation.Nullable;
+
+import com.baseflow.geolocator.location.LocationServiceStatusReceiver;
+
+import io.flutter.plugin.common.BinaryMessenger;
+import io.flutter.plugin.common.EventChannel;
+
+public class LocationServiceHandlerImpl implements EventChannel.StreamHandler{
+
+    private static final String TAG = "LocationServiceHandler";
+
+    @Nullable private EventChannel channel;
+    @Nullable private Activity activity;
+    @Nullable private Context context;
+    @Nullable private LocationServiceStatusReceiver receiver;
+
+
+    void startListening(Context context, BinaryMessenger messenger){
+        if(channel != null){
+            Log.w(TAG, "Setting a event call handler before the last was disposed.");
+            stopListening();
+        }
+        channel = new EventChannel(messenger, "flutter.baseflow.com/geolocator_service_updates");
+        channel.setStreamHandler(this);
+        this.context = context;
+    }
+
+    void stopListening(){
+        if(channel == null){
+            return;
+        }
+        channel.setStreamHandler(null);
+        channel = null;
+    }
+
+    void setActivity(@Nullable Activity activity) { this.activity = activity;}
+
+    @Override
+    public void onListen(Object arguments, EventChannel.EventSink events) {
+        IntentFilter filter = new IntentFilter(LocationManager.PROVIDERS_CHANGED_ACTION);
+        filter.addAction(Intent.ACTION_PROVIDER_CHANGED);
+        this.receiver = new LocationServiceStatusReceiver(arguments, events);
+        if(activity == null) return;
+        activity.registerReceiver(this.receiver, filter);
+    }
+
+    @Override
+    public void onCancel(Object arguments) {
+        activity.unregisterReceiver(this.receiver);
+    }
+
+}
--- a/geolocator/android/src/main/java/com/baseflow/geolocator/LocationServiceHandlerImpl.java
+++ b/geolocator/android/src/main/java/com/baseflow/geolocator/LocationServiceHandlerImpl.java
@@ -46,6 +46,10 @@ public class LocationServiceHandlerImpl implements EventChannel.StreamHandler{
 
     @Override
     public void onListen(Object arguments, EventChannel.EventSink events) {
+        if (activity == null) {
+            return;
+        }
+
         IntentFilter filter = new IntentFilter(LocationManager.PROVIDERS_CHANGED_ACTION);
         filter.addAction(Intent.ACTION_PROVIDER_CHANGED);
         this.receiver = new LocationServiceStatusReceiver(events);
--- a/geolocator/android/src/main/java/com/baseflow/geolocator/LocationServiceHandlerImpl.java
+++ b/geolocator/android/src/main/java/com/baseflow/geolocator/LocationServiceHandlerImpl.java
@@ -48,7 +48,7 @@ public class LocationServiceHandlerImpl implements EventChannel.StreamHandler{
     public void onListen(Object arguments, EventChannel.EventSink events) {
         IntentFilter filter = new IntentFilter(LocationManager.PROVIDERS_CHANGED_ACTION);
         filter.addAction(Intent.ACTION_PROVIDER_CHANGED);
-        this.receiver = new LocationServiceStatusReceiver(arguments, events);
+        this.receiver = new LocationServiceStatusReceiver(events);
         if(activity == null) return;
         activity.registerReceiver(this.receiver, filter);
     }
--- a/geolocator/android/src/main/java/com/baseflow/geolocator/location/LocationServiceStatusReceiver.java
+++ b/geolocator/android/src/main/java/com/baseflow/geolocator/location/LocationServiceStatusReceiver.java
@@ -29,7 +29,7 @@ public class LocationServiceStatusReceiver extends BroadcastReceiver {
             boolean isNetworkEnabled = locationManager.isProviderEnabled(LocationManager.NETWORK_PROVIDER);
 
 
-            if (isGpsEnabled) {
+            if (isGpsEnabled || isNetworkEnabled) {
                 /*
                 It may occur that the broadcastreceiver receives several events of the same type,
                 that's why u should check whether or not the event is fired more then one time.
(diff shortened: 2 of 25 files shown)


geolocator/example/lib/main.dart
```diff
@@ -201,10 +252,7 @@ class _GeolocatorWidgetState extends State<GeolocatorWidget> {
   }
 }
 
-enum _PositionItemType {
-  permission,
-  position,
-}
+enum _PositionItemType { permission, position, locationservicestatus }
```
mvanbeusekom, line 255: Dart style guide suggests to use lower camel casing for enum values. Meaning the initial letter of a word is written in lower case, and initial letter of sub-sequent words are written with in upper case. Also add a comma after the last enum so formatting will put each element on its own line:
```suggestion
enum _PositionItemType { 
  permission,
  position,
  locationServiceStatus, 
}
```

geolocator/lib/geolocator.dart
```diff
@@ -110,6 +110,7 @@ class Geolocator {
   /// location when the user denied access.
   /// Throws a [LocationServiceDisabledException] when the user allowed access,
   /// but the location services of the device are disabled.
+
```
mvanbeusekom, line 113: nit: redundant empty line which should be removed.
```suggestion
```

geolocator/ios/Classes/Utils/ServiceStatus.h
```diff
@@ -0,0 +1,22 @@
...
+//
+//  Created by Floris Smit on 03/06/2021.
+//
+
+#ifndef ServiceStatus_h
+#define ServiceStatus_h
+
+@interface ServiceStatusEnum : NSObject {
+    
+}
```
mvanbeusekom, line 13: This interface definition can be removed.
```suggestion
```

mvanbeusekom (changes requested):
Hi @florissmit1, 

Good start, I made some remarks in the Dart and Java implementation which in overal look good but can have some small improvements. 

Regarding the Objective-C implementation, I think we need to split the logic of the `FlutterStreamHandler` into two different classes (similar to the Java implementation). That means we need to create a separate `PositionStreamHandler` and a `LocationServiceStreamHandler` both implementing the `FlutterStreamHandler` class. This would solve the problem of the current `onListenWithArguments` method where two streams are mixed on different channels.

Now I can imagine that this sounds a bit daunting, so if you'd like I can help and move out the current stream handler implementation (which takes care of the position updates) into its own class so you have an example you can look at and copy for the location service update stream handler. Please let me know if this is something you would like and I will get started on it tomorrow morning.

In the meantime you can go through the other comments and make the necessary changes.

aytunch:
Sorry if this is not related but is there a `locationservice` concept in iOS? I was using a separate package to get this stream and it was only for Android. 
So when this is merged will it work for both Android and iOS? 
Can't wait this to be merged to geolocator so we use one less package :)

mvanbeusekom:
> Sorry if this is not related but is there a `locationservice` concept in iOS? I was using a separate package to get this stream and it was only for Android.
> So when this is merged will it work for both Android and iOS?
> Can't wait this to be merged to geolocator so we use one less package :)

Hi @aytunch, yes this update will also work on iOS. We are close on finishing the feature, just a little more refactoring on the iOS side, so hang on just a little bit longer ;).

geolocator/android/src/main/java/com/baseflow/geolocator/LocationServiceHandlerImpl.java
```diff
@@ -0,0 +1,61 @@
...
+
+    void setActivity(@Nullable Activity activity) { this.activity = activity;}
+
+    @Override
+    public void onListen(Object arguments, EventChannel.EventSink events) {
+        IntentFilter filter = new IntentFilter(LocationManager.PROVIDERS_CHANGED_ACTION);
+        filter.addAction(Intent.ACTION_PROVIDER_CHANGED);
+        this.receiver = new LocationServiceStatusReceiver(events);
+        if(activity == null) return;
+        activity.registerReceiver(this.receiver, filter);
```
mvanbeusekom, line 53: Move this check to the top of the method. Now you are executing quite some code that is unnecessary when the `activity == null`.

Two other small nits:
1. We try to follow the Google Java Style guide which recommends always to uses braces for your `if` statement (see suggestion);
2. You can remove the redundant `this.` syntax (so the code is consistent with the other code in the project).

```suggestion
        if(activity == null) { 
            return;
        }
        
        IntentFilter filter = new IntentFilter(LocationManager.PROVIDERS_CHANGED_ACTION);
        filter.addAction(Intent.ACTION_PROVIDER_CHANGED);
       receiver = new LocationServiceStatusReceiver(events);
        activity.registerReceiver(receiver, filter);
```

geolocator/android/src/main/java/com/baseflow/geolocator/location/LocationServiceStatusReceiver.java
```diff
@@ -0,0 +1,49 @@
...
+            } else {
+                if (isEnabled) {
+                    isEnabled = false;
+                    events.success(ServiceStatus.disabled.ordinal());
+                }
+
+            }
+        }
+    }
+}
```
mvanbeusekom, line 49: nit: it is considered to always end a file with a new line. In some cases this might lead to problems when using code automation tools not being able to parse the file correctly if the new line is missing. Although it probably won't cause any problems it is still good practice to add the line.

```suggestion
}

```

geolocator/ios/Classes/GeolocatorPlugin.m
```diff
@@ -24,11 +25,18 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar>*)registrar {
                                          eventChannelWithName:@"flutter.baseflow.com/geolocator_updates"
                                          binaryMessenger:registrar.messenger];
     
+    FlutterEventChannel *locationServiceUpdatesEventChannel = [FlutterEventChannel eventChannelWithName:@"flutter.baseflow.com/geolocator_service_updates" binaryMessenger:registrar.messenger];
+    
+
```
mvanbeusekom, line 30: nit: remove redundant empty line:

```suggestion
```

geolocator/ios/Classes/GeolocatorPlugin.m
```diff
@@ -24,11 +25,18 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar>*)registrar {
...
     
+    FlutterEventChannel *locationServiceUpdatesEventChannel = [FlutterEventChannel eventChannelWithName:@"flutter.baseflow.com/geolocator_service_updates" binaryMessenger:registrar.messenger];
+    
+    
     GeolocatorPlugin *instance = [[GeolocatorPlugin alloc] init];
     [registrar addMethodCallDelegate:instance channel:methodChannel];
   
     PositionStreamHandler *positionStreamHandler = [[PositionStreamHandler alloc] initWithGeolocationHandler:instance.geolocationHandler
                                                                                            PermissionHandler:instance.permissionHandler];
+
```
mvanbeusekom, line 36: nit: remove redundant empty line
```suggestion
```

geolocator/ios/Classes/Handlers/LocationServiceStreamHandler.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef LocationServiceStreamHandler_h
+#define LocationServiceStreamHandler_h
+
+#import <Flutter/Flutter.h>
+
+@interface LocationServiceStreamHandler : NSObject<FlutterStreamHandler, CLLocationManagerDelegate>
+
+@end
+
+
```
mvanbeusekom, line 17: nit: remove redundant empty line

```suggestion
```

geolocator/ios/Classes/Handlers/LocationServiceStreamHandler.m
```diff
@@ -0,0 +1,47 @@
...
+
+@interface LocationServiceStreamHandler()
+@property (strong, nonatomic) CLLocationManager *locationManager;
+@end
+
+@implementation LocationServiceStreamHandler {
+
+FlutterEventSink _eventSink;
+
+}
```
mvanbeusekom, line 22: nit: remove redundant empty lines (code should be concise, see also [Objective-C code style](https://google.github.io/styleguide/objcguide.html) guidelines):
```suggestion
@implementation LocationServiceStreamHandler {
  FlutterEventSink _eventSink;
}
```

geolocator/ios/Classes/Handlers/LocationServiceStreamHandler.m
```diff
@@ -0,0 +1,47 @@
...
+    _eventSink = nil;
+    return nil;
+}
+
+- (FlutterError * _Nullable)onListenWithArguments:(id _Nullable)arguments eventSink:(nonnull FlutterEventSink)events {
+    _eventSink = events;
+            if(self.locationManager == nil){
+                self.locationManager = [[CLLocationManager alloc] init];
+                self.locationManager.delegate = self;
+            }
```
mvanbeusekom, line 35: nit: Correct the indentation and add spaces around the if:

```suggestion
    if (self.locationManager == nil) {
        self.locationManager = [[CLLocationManager alloc] init];
        self.locationManager.delegate = self;
    }
```

geolocator/ios/Classes/Handlers/LocationServiceStreamHandler.m
```diff
@@ -0,0 +1,47 @@
...
+    return nil;
+}
+
+- (void)locationManager:(CLLocationManager *)manager didChangeAuthorizationStatus:(CLAuthorizationStatus)status{
+    if([CLLocationManager locationServicesEnabled]){
```
mvanbeusekom, line 40: nit: add spaces around the `if` statement:

```suggestion
    if ([CLLocationManager locationServicesEnabled]) {
```

mvanbeusekom (changes requested):
The code implementation looks good, but I left a whole bunch of comments regarding the code style which should be fixed.

mvanbeusekom (approved):
LGTM

mvanbeusekom:
@aytunch we just released version 7.1.0 which contains the new `Geolocator.getServiceStatusStream()` method which will allow you to listen for updates regarding the location service.

Please let us know if you have any questions or encounter any problems.

aytunch:
@mvanbeusekom thanks for the mention. I am so glad `getServiceStatusStream` is now in `geolocator`. I will be replacing the stream I used from `location_permissions` in to `geolocator` and test my app. I will let you know if any problems occur.

PR #769, opened 2021-07-02
Clarified docs getLocationAccuracy and added enum value for Android devices

florissmit1:
…or android to the LocationAccuracyStatus enum

Docs update / small implementation change

Currently when Android users execute the `getLocationAccuracy()` method a PluginException will be thrown. Also, the code documentation did not mention about the iOS only support.

When Android users execute the `getLocationAccuracy()` method now, a LocationAccuracyStatus.unknown will be returned, since Android does not support Approximate location yet. Also the documentation has been clarified.

No

Does not apply

Does not apply

### DIFF
--- a/geolocator_platform_interface/CHANGELOG.md
+++ b/geolocator_platform_interface/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.2.1
+
+- Documentation `getLocationAccuracy()` method clarified
+- On Android devices, `getLocationAccuracy()` now returns a LocationAccuracyStatus.unknown value
+
 ## 2.2.0
 
 - Added the possibility to query for the LocationAccuracyStatus on devices running iOS 14.0 and higher.
--- a/geolocator_platform_interface/CHANGELOG.md
+++ b/geolocator_platform_interface/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 2.2.1
 
-- Documentation `getLocationAccuracy()` method clarified
-- On Android devices, `getLocationAccuracy()` now returns a LocationAccuracyStatus.unknown value
+- Documentation `getLocationAccuracy()` method clarified.
+- Extended the `LocationAccuracyStatus` enum with a `LocationAccuracyStatus.unknown` value, which can be used by platforms that don't support location accuracy features.
 
 ## 2.2.0
 
--- a/geolocator_platform_interface/lib/src/geolocator_platform_interface.dart
+++ b/geolocator_platform_interface/lib/src/geolocator_platform_interface.dart
@@ -165,8 +165,8 @@ abstract class GeolocatorPlatform extends PlatformInterface {
   /// [LocationAccuracyStatus.reduced] will be returned, if the user gave
   /// permission for precise/full accuracy location, [LocationAccuracyStatus.precise]
   /// will be returned.
-  /// When executing the method on Android device's,
-  /// [LocationAccuracyStatus.unknown] will be returned.
+  /// When executing the method on platforms that don't support location
+  /// accuracy features [LocationAccuracyStatus.unknown] should be returned.
   Future<LocationAccuracyStatus> getLocationAccuracy() async {
     throw UnimplementedError('getLocationAccuracy() has not been implemented.');
   }
--- a/geolocator_platform_interface/lib/src/geolocator_platform_interface.dart
+++ b/geolocator_platform_interface/lib/src/geolocator_platform_interface.dart
@@ -159,12 +159,14 @@ abstract class GeolocatorPlatform extends PlatformInterface {
     throw UnimplementedError('getPositionStream() has not been implemented.');
   }
 
-  /// Returns a [Future] containing a [LocationAccuracyStatus].
+  /// Returns a [Future] containing a [LocationAccuracyStatus] (iOS only).
   ///
   /// When on iOS the user has given permission for approximate location,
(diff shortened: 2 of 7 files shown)


geolocator_platform_interface/lib/src/implementations/method_channel_geolocator.dart
```diff
@@ -92,6 +93,9 @@ class MethodChannelGeolocator extends GeolocatorPlatform {
 
   @override
   Future<LocationAccuracyStatus> getLocationAccuracy() async {
+    if (defaultTargetPlatform == TargetPlatform.iOS) {
+      return LocationAccuracyStatus.unknown;
+    }
```
mvanbeusekom, line 98: This code should be part of the `geolocator` package and not the `geolocator_platform_interface`.  The `geolocator_platoform_interface` package should always be platform agnostic (as in it should not know anything about platforms or platform specific code).

Also the condition is not correct. It will now always return `LocationAccuracyStatus.unknown` when the platform IS iOS.

geolocator_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.2.1
+
+- Documentation `getLocationAccuracy()` method clarified
```
mvanbeusekom, line 3: Missing period:
```suggestion
- Documentation `getLocationAccuracy()` method clarified.
```

geolocator_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+- On Android devices, `getLocationAccuracy()` now returns a LocationAccuracyStatus.unknown value
```
mvanbeusekom, line 4: The `geolocator_platform_interface` should be platform agnostic (not know about platform compatibility) and therefore we should not add the implementation to return a different state on Android in this package (see also later comment). This makes this release note incorrect maybe update it to something like this:  

```suggestion
- Extended the `LocationAccuracyStatus` enum with a `LocationAccuracyStatus.unknown` value, which can be used by platforms that don't support location accuracy features.
```

geolocator_platform_interface/lib/src/geolocator_platform_interface.dart
```diff
@@ -159,12 +159,14 @@ abstract class GeolocatorPlatform extends PlatformInterface {
...
 
-  /// Returns a [Future] containing a [LocationAccuracyStatus].
+  /// Returns a [Future] containing a [LocationAccuracyStatus] (iOS only).
   ///
   /// When on iOS the user has given permission for approximate location,
   /// [LocationAccuracyStatus.reduced] will be returned, if the user gave
   /// permission for precise/full accuracy location, [LocationAccuracyStatus.precise]
   /// will be returned.
+  /// When executing the method on Android device's,
+  /// [LocationAccuracyStatus.unknown] will be returned.
```
mvanbeusekom, line 169: The geolocator also supports the `Web` platform and hopefully soon also others like Linux, MacOS, Windows and embedded systems. So this documentation is not correct maybe update it to:

 ```suggestion
  /// When executing the method on platforms that don't support location 
  /// accuracy features [LocationAccuracyStatus.unknown] should be returned.
```

File: geolocator_platform_interface/test/src/implementations/method_channel_geolocator_test.dart [hunk 5]
```diff
@@ -154,24 +153,39 @@ void main() {
...
+          result: 2,
+        );
 
+        // Act
+        final locationAccuracy =
+            await MethodChannelGeolocator().getLocationAccuracy();
+
+        // Assert
+        expect(locationAccuracy, LocationAccuracyStatus.unknown);
+      });
```
mvanbeusekom, line 188: This unit-test can be removed as the unknown implementation should not be part of the `geolocator_platform_interface`.

geolocator_platform_interface/test/src/implementations/method_channel_geolocator_test.dart
```diff
@@ -154,24 +153,39 @@ void main() {
...
-      test(
-          'Should receive reduced accuracy if Location Accuracy is reduced',
-              () async {
-            // Arrange
-            MethodChannelMock(
-              channelName: 'flutter.baseflow.com/geolocator',
-              method: 'getLocationAccuracy',
-              result: 1,
-            );
+      test('Should receive reduced accuracy if Location Accuracy is reduced',
```
mvanbeusekom, line 156: This test has the same name as the the test on line 139 which is not allowed and should have a different name explaining the differences.

mvanbeusekom (changes requested):

geolocator_platform_interface/lib/src/implementations/method_channel_geolocator.dart
```diff
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:flutter/foundation.dart';
```
mvanbeusekom, line 3: nit: I think this import has become obsolete and can be removed.
```suggestion
```

mvanbeusekom (changes requested):
One small last change.

mvanbeusekom (approved):
LGTM

helixbooks:
Hi there! This is still broken for me on v7.4.0.

When running `getLocationAccuracy()` on an Android device, a `MissingPluginException` is still being thrown.

Issue #780, opened 2021-07-06
Location Services indicator never disappears


gridfriday:
If I get the device's location using Geolocator.getPositionStream() and then later cancel the subscription, the location services indicator in the status bar never gets removed. Nothing short of force quitting the app will remove the indicator.

`late StreamSubscription<Position> positionStream;`

`void getPosition() {
   positionStream = Geolocator.getPositionStream().listen((Position pos) {
      ...
   });
}`

`void stopLocation() {
   positionStream.cancel();
}`

Is this expected behavior?

If I instead use Geolocator.getCurrentPosition(), the indicator will disappear after a few seconds. But I want continuous user location updates.

I'm using Geolocator 7.2.0+1, Flutter 2.2.0 and seeing the issue on the iOS 14.4 simulator and a real device running 10.2.

MrBorsti:
The cause seems to be the `controller.stream.asBroadcastStream()` in the `_wrapStream`. If the broadcast stream is canceled the underlying location subscription is not.
Changing it to `controller.stream.asBroadcastStream(onCancel: (sub) => sub.cancel());` or even change the whole stream to be a broadcast from `StreamController<dynamic>(...` to `StreamController<dynamic>.broadcast(...` will close everything and the indicator disappears.

What is the reason for this?

incloudss:
Do anyone know maybe if getting position stream with given configuration, then cancelling it, and then again getting position stream, but with different configuration(different time interval for example) will respect new configuration? Because from my quick investigaton, it will probably return old stream, and therefore the configuration will be not changed. It would also fits the theory with not disappearing indicator.

huangkaichang:
Remove stream while cancelling the subscription, fixed this problem for me.

cto-leaps:
Hi guys and thank you so much for this great library!
Any news on when this fix will be released?
many 🙏 in advance

mvanbeusekom:
@gridfriday, @MrBorsti, @incloudss, @cto-leaps, this issue is solved with version 2.3.1 of the geolocator_platform_interface package (which the geolocator package depends on).

The current version (7.2.0+1) of the geolocator plugin depends the geolocator_platform_interface using the following version `^2.2.0` constraint. So running `flutter pub upgrade` should resolve the issue. 

Today I will release a new version of the geolocator plugin package which will also update the constraint of the geolocator_platform_interface which should also remove the need of running `flutter pub upgrade`.

mvanbeusekom:
[...]

PR #782, opened 2021-07-08
Fix unsubscribe not removing position stream

stsc3000:
Bug fix

When using `getPositionStream`, the original position stream is never cancelled despite all listeners of `getPositionStream` being cancelled,
thus the native position subscription is never stopped.

Unfortunately, _from what I understand_,  the accompanying [test](https://github.com/Baseflow/flutter-geolocator/blob/master/geolocator_platform_interface/test/src/implementations/method_channel_geolocator_test.dart#L662-L682) has a faulty expectation (`==` instead of `!=`) and also implies that given options have any influence on reusing the `_positionStream` (which might be a different problem by itself). 

Cancelling all listeners of the position broadcast stream cancels the position stream as well.
Accompanying test is a bit more extensive and does no longer imply that given options influence reuse of `_positionStream`.

No

Use the example app and adapt to use cancel instead of pause.

Issue #780

### DIFF
--- a/geolocator_platform_interface/CHANGELOG.md
+++ b/geolocator_platform_interface/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.3.1
+
+- Solves a bug which resulted in an issue when closing the position stream.
+
 ## 2.3.0
 
 - Added the possibility to request temporary Precise Accuracy on iOS 14+ devices.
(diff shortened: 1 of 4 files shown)


mvanbeusekom (approved):
LGTM, thank you very much for providing this fix

PR #791, opened 2021-07-22
Add pausesLocationUpdatesAutomatically parameter

cdfq152313:
Feature.

Location update may be pasue in iOS when user stop for a while.

Add pausesLocationUpdatesAutomatically parameter.

No.

https://developer.apple.com/documentation/corelocation/cllocationmanager/1620553-pauseslocationupdatesautomatical?language=objc

- [v] I made sure all projects build.
- [v] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning).
- [v] I updated CHANGELOG.md to add a description of the change.
- [v] I followed the style guide lines ([code style guide](https://github.com/Baseflow/flutter-geolocator/blob/master/CONTRIBUTING.md)).
- [v] I updated the relevant documentation.
- [v] I rebased onto current `master`.

### DIFF
--- a/geolocator/lib/geolocator.dart
+++ b/geolocator/lib/geolocator.dart
@@ -114,6 +114,7 @@ class Geolocator {
     LocationAccuracy desiredAccuracy = LocationAccuracy.best,
     int distanceFilter = 0,
     bool forceAndroidLocationManager = false,
+    bool iosPausesLocationUpdatesAutomatically = false,
     Duration? intervalDuration,
     Duration? timeLimit,
   }) =>
@@ -121,6 +122,7 @@ class Geolocator {
         desiredAccuracy: desiredAccuracy,
         distanceFilter: distanceFilter,
         forceAndroidLocationManager: forceAndroidLocationManager,
+        iosPausesLocationUpdatesAutomatically: iosPausesLocationUpdatesAutomatically,
         timeInterval: intervalDuration?.inMilliseconds ?? 0,
         timeLimit: timeLimit,
       );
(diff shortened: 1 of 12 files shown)


mvanbeusekom (changes requested):
Hi @cdfq152313, 

Thank you for this pull request, this looks like an interesting feature that would be nice to support. There are a few issues with the current implementation though, which would be awesome if you could solve:

1. According to Apple's documentation the `CLLocationManager.setPausesLocationUpdatesAutomatically` method goes hand in hand with the `CLLocationManager.activityType` property. So setting the `setPausesLocationUpdatesAutomatically` isn't really helpful without also setting the `activityType`. Before we start adding an additional parameter to the list, I think it would be good to create a separate `IosSpecificSettings` class which we send as a `Map<String, Object>` over to the native side. This class can contain the `activityType` and the `pausesLocationUpdatesAutomatically` settings and can also be used in the future to add more iOS specific settings. To make it really flexible it would be awesome if the `geolocator_platform_interface` defines a `abstract PlatformSpecificSettings` class that is implemented by the `IosSpecificSettings` class and defines a `toMap` method that is used to convert the different properties into the `Map<String, Object>` datatype used to transfer the information to the native side. This way the method signature could simply be:

```
Geolocator.getPositionStream({
   LocationAccuracy desiredAccuracy,
   int distanceFilter,
   Duration? intervalDuration,
   Duration? timeLimit,
   bool forceAndroidLocationManager,
   PlatformSpecificSettings? platformSettings,
}) {
  ...
}
```
2. Please create separate pull requests for the changes to the `geolocator_platform_interface` package and the `geolocator` package. This will allow us to publish the `geolocator_platform_interface` package first and afterwards also update the dependency in the `geolocator` package;
3. Please have a look at the errors reported by the CI environment. At the moment the several tasks are failing.

PR #802, opened 2021-07-30
Ios location options geolocator_apple package

florissmit1:
Feature

Added the ActivityType enum/property and the pauseLocationUpdatesAutomatically property to the geolocator_apple package

No

Does not apply

Does not apply


geolocator_apple/lib/ios_options.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:geolocator_platform_interface/geolocator_platform_interface.dart';
+import 'activity_type.dart';
+
+/// Represents different iOS specific settings with which you can set a value
+/// other then the default value of the setting.
+class IosOptions extends LocationOptions {
```
mvanbeusekom, line 6: I think we agreed to call this `AppleSettings` (not `IosOptions`). Also the base class should be the `LocationSettings` class (`LocationOptions` doesn't exist anymore in your new version of the `geoloator_platform_interface` package).

mvanbeusekom (changes requested):
Would be nice if we make a barrel file which exports all public types, so consumers don't have to import a lot of different files:

`lib/geolocator_apple.dart` barrel file:

```dart
export 'src/types/activity_type.dart';
export 'src/types/apple_settings.dart';
```

This means you need to move the `activity_type.dart` and `apple_settings.dart` into the `lib/src/types` folder.

geolocator_apple/lib/ios_options.dart
florissmit1: You're right, my bad.

geolocator_apple/ios/Classes/Handlers/GeolocationHandler.m
```diff
@@ -34,6 +36,14 @@ - (void)startListeningWithDesiredAccuracy:(CLLocationAccuracy)desiredAccuracy
     CLLocationManager *locationManager = self.locationManager;
     locationManager.desiredAccuracy = desiredAccuracy;
     locationManager.distanceFilter = distanceFilter == 0 ? kCLDistanceFilterNone : distanceFilter;
+    locationManager.activityType = activityType;
+    
+    if (!(pauseLocationUpdatesAutomatically == nil)) {
+        if ([pauseLocationUpdatesAutomatically boolValue]) {
+            locationManager.pausesLocationUpdatesAutomatically = true;
+        }
+    }
```
mvanbeusekom, line 45: This statement can be simplified to (as it is already passed as a boolean):

```suggestion
    locationManager.pausesLocationUpdatesAutomatically = pauseLocationUpdatesAutomatically;

```

geolocator_apple/ios/Classes/GeolocatorPlugin.m
```diff
@@ -119,10 +120,14 @@ - (void)onGetCurrentPositionWithArguments:(id _Nullable)arguments
     
     CLLocationAccuracy accuracy = [LocationAccuracyMapper toCLLocationAccuracy:(NSNumber *)arguments[@"accuracy"]];
     CLLocationDistance distanceFilter = [LocationDistanceMapper toCLLocationDistance:(NSNumber *)arguments[@"distanceFilter"]];
+      NSNumber* pauseLocationUpdatesAutomatically = [NSNumber numberWithBool:(bool)arguments[@"pauseLocationUpdatesAutomatically"]];
```
mvanbeusekom, line 123: We should try to parse data coming from Dart as early as possible and work with strong types in the rest of the code. In this case the `pauseLocationUpdatesAutomatically` parameter is supposed to be a boolean, however this is transported from Dart to Objective-C as `NSNumber` pointer (because theoretically it can be `null`, see also the [Flutter documentation](https://flutter.dev/docs/development/platform-integration/platform-channels#codec)). 

This means we need to convert it back to a boolean as soon as possible, and since we determined that a `null` value should be treated as `false` we can do that here. So first lets a reference to the `NSNumber` pointer:

```suggestion
      NSNumber* pauseLocationUpdatesAutomatically = arguments[@"pauseLocationUpdatesAutomatically"];
```

Notice that I removed the `(bool)` cast, this will give a problem if someone sends `nil`.

And when we pass it to the `startListeningWithDesiredAccuracy` method we convert it to a boolean (see next comment).

geolocator_apple/ios/Classes/GeolocatorPlugin.m
```diff
@@ -119,10 +120,14 @@ - (void)onGetCurrentPositionWithArguments:(id _Nullable)arguments
...
+      CLActivityType activityType = [ActivityTypeMapper toCLActivityType:(NSNumber *)arguments[@"activityType"]];
     GeolocationHandler *geolocationHandler = [[GeolocationHandler alloc] init];
     
     [geolocationHandler startListeningWithDesiredAccuracy:accuracy
                                            distanceFilter:distanceFilter
+                        pauseLocationUpdatesAutomatically:pauseLocationUpdatesAutomatically
```
mvanbeusekom, line 129: Don't pass a `NSNumber` pointer but the strongly type boolean value:
```suggestion
                        pauseLocationUpdatesAutomatically:pauseLocationUpdatesAutomatically && [pauseLocationUpdatesAutomatically boolValue]
```

geolocator_apple/ios/Classes/Handlers/GeolocationHandler.h
```diff
@@ -20,6 +20,8 @@ typedef void (^GeolocatorResult)(CLLocation *_Nullable location);
 
 - (void)startListeningWithDesiredAccuracy:(CLLocationAccuracy)desiredAccuracy
                            distanceFilter:(CLLocationDistance)distanceFilter
+pauseLocationUpdatesAutomatically:(NSNumber* _Nullable)pauseLocationUpdatesAutomatically
```
mvanbeusekom, line 23: Don't accept the `NSNumber` type here if we know we need a boolean (this will force calling code to deal with conversions):
```suggestion
pauseLocationUpdatesAutomatically:(BOOL)pauseLocationUpdatesAutomatically
```

geolocator_apple/ios/Classes/Handlers/PositionStreamHandler.m
```diff
@@ -55,9 +56,14 @@ - (FlutterError *_Nullable)onListenWithArguments:(id _Nullable)arguments eventSi
       
       CLLocationAccuracy accuracy = [LocationAccuracyMapper toCLLocationAccuracy:(NSNumber *)arguments[@"accuracy"]];
       CLLocationDistance distanceFilter = [LocationDistanceMapper toCLLocationDistance:(NSNumber *)arguments[@"distanceFilter"]];
+      NSNumber* pauseLocationUpdatesAutomatically = [NSNumber numberWithBool:(bool)arguments[@"pauseLocationUpdatesAutomatically"]];
```
mvanbeusekom, line 59: We should try to parse data coming from Dart as early as possible and work with strong types in the rest of the code. In this case the `pauseLocationUpdatesAutomatically` parameter is supposed to be a boolean, however this is transported from Dart to Objective-C as `NSNumber` pointer (because theoretically it can be `null`, see also the [Flutter documentation](https://flutter.dev/docs/development/platform-integration/platform-channels#codec)). 

This means we need to convert it back to a boolean as soon as possible, and since we determined that a `null` value should be treated as `false` we can do that here. So first lets a reference to the `NSNumber` pointer:

```suggestion
      NSNumber* pauseLocationUpdatesAutomatically = arguments[@"pauseLocationUpdatesAutomatically"];
```

Notice that I removed the `(bool)` cast, this will give a problem if someone sends `nil`.

And when we pass it to the `startListeningWithDesiredAccuracy` method we convert it to a boolean (see next comment).

geolocator_apple/ios/Classes/Handlers/PositionStreamHandler.m
```diff
@@ -55,9 +56,14 @@ - (FlutterError *_Nullable)onListenWithArguments:(id _Nullable)arguments eventSi
...
+      CLActivityType activityType = [ActivityTypeMapper toCLActivityType:(NSNumber *)arguments[@"activityType"]];
       
       [[weakSelf geolocationHandler] startListeningWithDesiredAccuracy:accuracy
                                                     distanceFilter:distanceFilter
+                                     pauseLocationUpdatesAutomatically:pauseLocationUpdatesAutomatically
```
mvanbeusekom, line 64: Don't pass a `NSNumber` pointer but the strongly type boolean value:
```suggestion
                        pauseLocationUpdatesAutomatically:pauseLocationUpdatesAutomatically && [pauseLocationUpdatesAutomatically boolValue]
```

geolocator_apple/lib/src/types/activity_type.dart
```diff
@@ -0,0 +1,21 @@
+/// Represents the possible iOS activity types
```
mvanbeusekom, line 1: nit: missing period.

```suggestion
/// Represents the possible iOS activity types.
```

geolocator_apple/lib/src/types/apple_settings.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:geolocator_platform_interface/geolocator_platform_interface.dart';
+import 'activity_type.dart';
+
+/// Represents different iOS specific settings with which you can set a value
+/// other then the default value of the setting.
+class AppleSettings extends LocationSettings {
+  /// Initializes a new [IosSpecificSettings] instance with default values.
```
mvanbeusekom, line 7: Use correct type in comments:

```suggestion
  /// Initializes a new [AppleSettings] instance with default values.
```

geolocator_apple/pubspec.yaml
```diff
@@ -17,8 +20,9 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
-  
-  geolocator_platform_interface: ^2.3.2
+  # TODO: Path reference should be removed
```
mvanbeusekom, line 23: nit: It is very helpful for others to explain when the path reference should be removed, similar to the comment you added at the `publish_to` tag:

```suggestion
  # TODO: Path reference should be removed when the final version of the geolocator_platform_interface is published.
```

mvanbeusekom (changes requested):

mvanbeusekom:
@florissmit1 also the `geolocator_apple / Format` step failed indicating the formatting of the Dart code is invalid.

mvanbeusekom (approved):
LGTM

PR #803, opened 2021-07-30
Android location options added to the geolocator_android package

florissmit1:
Feature

Added the Android location options to the `lib` directory in the geolocator_android package. Moved the `forceAndroidLocationManager` property to the Android Location options instance.

No

Does not apply

Does not apply


geolocator_android/lib/android_options.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:geolocator_platform_interface/geolocator_platform_interface.dart';
+
+/// Represents different Android specific settings with which you can set a value
+/// other then the default value of the setting.
+class AndroidOptions extends LocationOptions {
```
mvanbeusekom, line 5: This should be `AndroidSettings` and extend the `LocationSettings` class from the geolocator_platform_interface package.

mvanbeusekom (changes requested):
Same as with the Apple version, it would be good to create a barrel file which exposes the new types so consumers only have to import one file.

geolocator_android/lib/src/types/android_settings.dart
```diff
@@ -0,0 +1,38 @@
...
+class AndroidSettings extends LocationSettings {
+  /// Initializes a new [AndroidSpecificSettings] instance with default values.
+  ///
+  /// The following default values are used:
+  /// - forceLocationManager: false
+  AndroidSettings({
+    this.forceLocationManager = false,
+    LocationAccuracy accuracy = LocationAccuracy.best,
+    int distanceFilter = 0,
+    Duration? intervalDuration,
```
mvanbeusekom, line 14: Shouldn't this also include the `timeLimit` parameter?

mvanbeusekom (changes requested):

geolocator_android/lib/src/types/android_settings.dart
florissmit1: U're right!

Issue #816, opened 2021-08-18
didChangeAppLifecycleState loop


p-shapovalov:
## 🐛 Bug Report

**didChangeAppLifecycleState** goes into a loop after location request. Only on some devices, but definitely reproduced on any xiaomi redmi note 9
It can be avoided by **forceAndroidLocationManager: true**

### Configuration

**Version:** 7.4.0

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

p-shapovalov:
Please check https://github.com/Baseflow/flutter-geolocator/issues/717#issuecomment-822654900, same behavior here with activity result cycle

hoanglm4:
same issue for me on Xiaomi device

mvanbeusekom:
Thanks to @hoanglm4 this issue should now be fixed. I will close the issue but feel free to leave a comment when you run into the issue again.

sed1ka:
this issue is exactly what happened to me. This issue also happen on Xiaomi Redmi 9T #1056

PR #834, opened 2021-09-15
fix. didChangeAppLifecycleState** goes into a loop when starting the …

hoanglm4:
Bug fix.

**didChangeAppLifecycleState** goes into a loop after location request.
Version: 7.6.0
Config: **forceAndroidLocationManager**: false
Platform: Android only
Device Model: Xiaomi Only

**didChangeAppLifecycleState** not loop after location request.

None

None

https://github.com/Baseflow/flutter-geolocator/issues/816

### DIFF
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/FusedLocationClient.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/FusedLocationClient.java
@@ -69,6 +69,13 @@ class FusedLocationClient implements LocationClient {
     return random.nextInt(1 << 16);
   }
 
+  @SuppressLint("MissingPermission")
+  private void requestPositionUpdates(LocationOptions locationOptions) {
+    LocationRequest locationRequest = buildLocationRequest(locationOptions);
+    fusedLocationProviderClient.requestLocationUpdates(
+            locationRequest, locationCallback, Looper.getMainLooper());
+  }
+
   @Override
   public void isLocationServiceEnabled(LocationServiceListener listener) {
     LocationServices.getSettingsClient(context)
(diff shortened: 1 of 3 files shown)


geolocator_android/android/src/main/java/com/baseflow/geolocator/location/FusedLocationClient.java
```diff
@@ -140,6 +149,12 @@ public void startPositionUpdates(
     LocationRequest locationRequest = buildLocationRequest(this.locationOptions);
     LocationSettingsRequest settingsRequest = buildLocationSettingsRequest(locationRequest);
 
+      if (forceRequest) {
+          fusedLocationProviderClient.requestLocationUpdates(
+                  locationRequest, locationCallback, Looper.getMainLooper());
+          return;
+      }
+
```
mvanbeusekom, line 157: We could extract this logic into a separate method and call it directly from the `onActivityResult` callback. That would take out this `if` statement and makes the code easier to understand. It will also reduce code duplication. I was thinking something like:

```java
private void requestPositionUpdates(LocationOptions locationOptions) {
    LocationRequest locationRequest = buildLocationRequest(locationOptions);
    fusedLocationProviderClient.requestLocationUpdates(
            locationRequest, locationCallback, Looper.getMainLooper());
}
```

Next we call this method on all locations where we currently call the `fusedLocationProviderClient.requestLocationUpdates()` method and in the `onActivityResult()` method where we currently call the `startPositionUpdates` method. This would remove the need for the `forceRequest` boolean and the additional if statement.

mvanbeusekom (changes requested):
Hi @hoanglm4, thank you for this pull-request. I have been looking a long time on how to fix this issue and all becomes clear now. 

I do have some feedback on how to improve on your proposed solution, I hope you could send an update. Also please update the version number of the `./geolocator_android/pubspec.yaml` file and add an entry in the `./geolocator_android/CHANGELOG.md` file with a short description of the fix you made. This will allow us to more easily publish the changes.

hoanglm4:
Hi @mvanbeusekom ,
Sorry too late. Could u please help me review again.

Thanks

mvanbeusekom (approved):
LGTM, thank you for the bug fix. I really appreciate it.

PR #850, opened 2021-09-30
Approximate Location support for Android 12

florissmit1:
Feature
Currently, the `geolocator` will mark Android 12 Approximate location requests as `Permission. denied`. This causes some apps running on Android to fail when trying to fetch the Location.

The issue mentioned above is resolved. Approximate location requests will now be marked as `Permission.granted`.

No

Use the example application

Does not apply

### DIFF
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
@@ -44,7 +44,7 @@ public class PermissionManager
         if (permissions.length == 1) {
             permissionStatus = ContextCompat.checkSelfPermission(context, permissions[0]);
         } else {
-            if(ContextCompat.checkSelfPermission(context, permissions[0]) == 0
+            if (ContextCompat.checkSelfPermission(context, permissions[0]) == 0
                     || ContextCompat.checkSelfPermission(context, permissions[1]) == 0) {
                 permissionStatus = PackageManager.PERMISSION_GRANTED;
             }
@@ -155,7 +155,7 @@ public class PermissionManager
             return false;
         }
 
-        if(requestedPermissions.length == 1) {
+        if (requestedPermissions.length == 1) {
             if (grantResults[requestedPermissionIndex] == PackageManager.PERMISSION_GRANTED) {
                 if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
                     int backgroundPermissionIndex =
@@ -176,7 +176,7 @@ public class PermissionManager
             }
         } else {
             if (grantResults[requestedPermissionIndex] == PackageManager.PERMISSION_GRANTED
-                    || grantResults[requestedPermissionIndex+1] == PackageManager.PERMISSION_GRANTED) {
+                    || grantResults[requestedPermissionIndex + 1] == PackageManager.PERMISSION_GRANTED) {
                 if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
                     int backgroundPermissionIndex =
                             indexOf(permissions, Manifest.permission.ACCESS_BACKGROUND_LOCATION);
@@ -191,7 +191,7 @@ public class PermissionManager
                 }
             } else {
                 if (activity != null && !ActivityCompat.shouldShowRequestPermissionRationale(activity, requestedPermissions[0])
-                    && !ActivityCompat.shouldShowRequestPermissionRationale(activity, requestedPermissions[1])) {
+                        && !ActivityCompat.shouldShowRequestPermissionRationale(activity, requestedPermissions[1])) {
                     permission = LocationPermission.deniedForever;
                 }
             }
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
@@ -19,172 +19,216 @@ import java.util.*;
 
 @SuppressWarnings("deprecation")
 public class PermissionManager
-    implements io.flutter.plugin.common.PluginRegistry.RequestPermissionsResultListener {
+        implements io.flutter.plugin.common.PluginRegistry.RequestPermissionsResultListener {
 
-  private static final int PERMISSION_REQUEST_CODE = 109;
+    private static final int PERMISSION_REQUEST_CODE = 109;
 
-  @Nullable private Activity activity;
-  @Nullable private ErrorCallback errorCallback;
-  @Nullable private PermissionResultCallback resultCallback;
+    @Nullable
+    private Activity activity;
+    @Nullable
+    private ErrorCallback errorCallback;
+    @Nullable
+    private PermissionResultCallback resultCallback;
 
-  public LocationPermission checkPermissionStatus(Context context, Activity activity)
-      throws PermissionUndefinedException {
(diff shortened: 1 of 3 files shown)


geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -142,22 +154,50 @@ public boolean onRequestPermissionsResult(
...
+          permission = LocationPermission.always;
+        }
+      } else {
+        if (activity != null
+            && !ActivityCompat.shouldShowRequestPermissionRationale(
+                activity, requestedPermissions[0])
+            && !ActivityCompat.shouldShowRequestPermissionRationale(
+                activity, requestedPermissions[1])) {
+          permission = LocationPermission.deniedForever;
+        }
```
mvanbeusekom, line 200: The code in this function can be heavily optimised by doing the following:

1. Don't access the permissions based on a fixed index, use a loop instead;
2. Add an extra method to determine if access to the `Manifest.permission.ACCESS_BACKGROUND_LOCATION` has been granted (this method could simply return a boolean indicating granted or not). Use this method i to determine if permissions should be `whileInUse` (background not allowed) or `always` (background allowed).

geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -29,16 +29,28 @@
...
+    if (permissions.length == 1) {
+      permissionStatus = ContextCompat.checkSelfPermission(context, permissions[0]);
+    } else {
+      if (ContextCompat.checkSelfPermission(context, permissions[0])
+              == PackageManager.PERMISSION_GRANTED
+          || ContextCompat.checkSelfPermission(context, permissions[1])
+              == PackageManager.PERMISSION_GRANTED) {
+        permissionStatus = PackageManager.PERMISSION_GRANTED;
+      }
+    }
```
mvanbeusekom, line 50: This code is rather sensitive for bugs because it is accessing elements in the `permissions` array directly based on a literal index (either 0 or 1).

The two main questions you should ask yourself are:
1. What if the array is empty and doesn't contain any elements at all?
2. What if additional permissions are introduced (a third, forth, etc...)? 

Have another look and see if you can solve these questions.

geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
mvanbeusekom, line 200: Same as in the `checkPermissionStatus` method, this code is very sensitive for error in case the array with permissions that should be requested is empty or contains more than two permissions (theoretically this function doesn't know how many permissions there are and it doesn't need to).

One additional tip, you can make this function easier by extracting the logic if background permissions are granted or not in to a separate method:

```java
private boolean hasBackgroundAccess(String[] permissions, int[] grantResults) {
  int backgroundPermissionIndex =
            indexOf(permissions, Manifest.permission.ACCESS_BACKGROUND_LOCATION);
  return backgroundPermissionIndex >= 0
              && grantResults[backgroundPermissionIndex] == PackageManager.PERMISSION_GRANTED;
} 
```

mvanbeusekom (changes requested):
The current implementation is quite fragile and contains room for improvement. I left comments on the places where the code can be improved. Has a look a try to fix them, if you get stuck let me know and I will try to explain more detailed.

One general remark (which goes for any programming language): accessing elements in an array or collection based on a literal index is (almost) always a bad thing.

geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -142,49 +153,79 @@ public boolean onRequestPermissionsResult(
...
     } else {
-      if (activity != null && !ActivityCompat.shouldShowRequestPermissionRationale(activity, requestedPermission)) {
-        permission = LocationPermission.deniedForever;
+      boolean shouldShowRationale = false;
+      for (String permission : requestedPermissions) {
+        if (ActivityCompat.shouldShowRequestPermissionRationale(activity, permission)) {
+          shouldShowRationale = true;
+          break;
+        }
+      }
```
mvanbeusekom, line 182: This could also be tracked in a separate field directly in the initial loop (starting at line 138). And only here check the `shouldShowRationale` variable to set the permission on `deniedForever` if needed.

mvanbeusekom (changes requested):
This looks a lot better, I do have some optimisation remarks.

geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -124,15 +131,19 @@ public boolean onRequestPermissionsResult(
...
-          "Geolocator",
-          "Location permissions not part of permissions send to onRequestPermissionsResult method.");
-      return false;
+    for (String permission : requestedPermissions) {
+      int requestedPermissionIndex = indexOf(permissions, permission);
+      if (requestedPermissionIndex < 0) {
+        Log.w(
+            "Geolocator",
+            "Location permissions not part of permissions send to onRequestPermissionsResult method.");
+        return false;
```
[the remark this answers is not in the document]
florissmit1: Not sure how to solve this correctly, just committed my solution, but I doubt if this is the correct solution.

geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -124,67 +132,89 @@ public boolean onRequestPermissionsResult(
...
+
+    for (String permission : requestedPermissions) {
+      int requestedPermissionIndex = indexOf(permissions, permission);
+      if (requestedPermissionIndex > 0) {
+          permissionsPartOfPermissionsResult = true;
+      }
+      if (requestedPermissionIndex == PackageManager.PERMISSION_GRANTED) {
+        grantedResult = PackageManager.PERMISSION_GRANTED;
       }
+      if (ActivityCompat.shouldShowRequestPermissionRationale(activity, permission)) {
```
mvanbeusekom, line 155: Only call the `shouldShowRequestPermissionRationale` when the `activity` is not `null`. Current implementation can result in a null reference exception.

When the `activity` is `null` the code should continue as if `shouldShowRequestPermissionRationale` returned `true`.

geolocator_android/android/src/main/java/com/baseflow/geolocator/permission/PermissionManager.java
```diff
@@ -124,67 +132,89 @@ public boolean onRequestPermissionsResult(
...
+    }
+
+    if (grantedResult == PackageManager.PERMISSION_GRANTED) {
+        locationPermission = (android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.Q || hasBackgroundAccess(permissions, grantResults))
+                ? LocationPermission.always
+                : LocationPermission.whileInUse;
     } else {
-      if (activity != null && !ActivityCompat.shouldShowRequestPermissionRationale(activity, requestedPermission)) {
-        permission = LocationPermission.deniedForever;
+      if (activity != null && !shouldShowRationale) {
```
mvanbeusekom, line 172: Checking if `activity` is `null` or not is not relevant here as it is not used in the context of this `if` statement or its body:

```suggestion
      if (!shouldShowRationale) {
```

mvanbeusekom (changes requested):
Please have a look at the findings. Current implementation contains several mistakes.

PR #853, opened 2021-10-04
Update stream cancelation callback

diegogarciar:
It seems that the cancel method is not being called, autocompletion declared it differently.

when canceling a stream created via `Geolocator.getPositionStream`, the cancel callback is not invoked on iOS

```dart
final stream = Geolocator.getPositionStream(
        desiredAccuracy: LocationAccuracy.best,
        intervalDuration: const Duration(seconds: 1),
      ).listen((event) { });
      stream.cancel();
```

the cancel callback is invoked on iOS

No

-Start a new `Geolocator.getPositionStream` and cancel the stream from a button
-Run from Xcode and  set a breakpoint inside cancelation callback
-debugger should stop at breakpoint, compare with master branch

### DIFF
--- a/geolocator_apple/apple/Classes/Handlers/PositionStreamHandler.m
+++ b/geolocator_apple/apple/Classes/Handlers/PositionStreamHandler.m
@@ -76,7 +76,7 @@
   return nil;
 }
 
-- (FlutterError *_Nullable)onCancelWithArguments:(id _Nullable)arguments {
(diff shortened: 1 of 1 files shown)


diegogarciar:
Should I bump the version to `1.2.2`?

mvanbeusekom:
@diegogarciar yes please that would be great. Please also include an entry in the geolocator_apple/CHANGELOG.md file.

Issue #821, opened 2021-08-26
[iOS] Not receiving location updates if location services are turned off and then back on


mikebryant71:
### Expected behavior
After subscribing to location updates (onLocationChanged stream) if the user turns off the location services, then back on, it should resume yielding new location updates.

### Reproduction steps

- Tap on "Play button" to start listening for position updates
- Go to native Settings/Privacy/Location Services 
- Turn off Location Services
- Turn on Location Services
- Return to Example app
- Change user's location via Simulator menu -> Features/Location
- The new position is not received. (onLocationChanged not receiving new data)

Video: https://streamable.com/cc1p5l

### Configuration
Use iOS baseflow example app provided. (Also reproducible in the app I'm developing)
Using iOS Simulator, reproducible on iOS 13.2 & 14.4. (likely on other versions of iOS too).

**Version:** 7.4.0

**Platform:** 
- [x] :iphone: iOS
- [ ] :robot: Android

mikebryant71:
I did some testing and did not find there to be any memory leaks, nor with energy usage. It seems like the CLLocationManager does a good job of that.

An exception, although an improvement, I think the current behaviour is a bit unexpected from the flutter side. If you're start listening to a stream, you expect it to be "active" until you call `cancel()`. Calling `cancel()` also stops listening to location updates so that means any memory concerns flutter side are manageable.
This also makes it easier for the flutter dev who wants to monitor location updates. They don't need to keep re subscribing to a stream based off the events of the service enabled stream which would also makes them independant.

florissmit1:
@JustChris This issue is solved, the PR is mentioned just above this comment. I will close this issue, if it's still not working like you would expect, please re-open this issue.

PR #896, opened 2021-11-18
Keep PositionStream alive when service status has been changed

JDDV:

Current behavior is that the PositionStream is cancelled when the `Location Service` or permission iOS devices is turned off. The behavior in this PR keeps the stream alive when the location service has been turned off and on again in the settings. 

Stops giving location updates when the location service has been turned off and on again in the settings.

Keeps the PositionStream alive when the location service has been turned off and on again in the settings.

No

Issue #821

### DIFF
--- a/geolocator_apple/apple/Classes/Handlers/GeolocationHandler.m
+++ b/geolocator_apple/apple/Classes/Handlers/GeolocationHandler.m
@@ -108,9 +108,7 @@
     if([error.domain isEqualToString:kCLErrorDomain] && error.code == kCLErrorLocationUnknown) {
         return;
     }
-    
-    
-    
+
     if (self.errorHandler) {
         self.errorHandler(GeolocatorErrorLocationUpdateFailure, error.localizedDescription);
     }
--- a/geolocator_apple/apple/Classes/Handlers/GeolocationHandler.m
+++ b/geolocator_apple/apple/Classes/Handlers/GeolocationHandler.m
@@ -109,7 +109,7 @@
         return;
     }
     
-    [self stopListening];
+    
     
     if (self.errorHandler) {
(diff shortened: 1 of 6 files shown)


geolocator_apple/apple/Classes/Handlers/GeolocationHandler.m
```diff
@@ -109,7 +109,7 @@ - (void)locationManager:(CLLocationManager *)manager
         return;
     }
     
-    [self stopListening];
+
```
mvanbeusekom, line 113: nit: let's remove all these empty lines and just keep one.

mvanbeusekom (changes requested):
Currently the build is failing due to the version number in `pubspec.yaml` being set to `version: NEXT`. This is something I didn't anticipate on and I fail it would be better to make sure the build runs correctly. 

@JDDV could you replace `NEXT` with the version to the next major number (this would be 2.0.0) in both the CHANGELOG.md and the pubspec.yaml file for the geolocator_apple package?

mvanbeusekom:
@JDDV code looks good but there is a unit test failing at the moment. Most likely because the stream is no longer cancelled and the unit test still expecting it is.

JDDV:
> @JDDV code looks good but there is a unit test failing at the moment. Most likely because the stream is no longer cancelled and the unit test still expecting it is.

I don't think that's the case here @mvanbeusekom, the test locally on my pc passes every single time.
Somehow the `TimeoutException` doesn't get thrown, I'll close and re-open the PR (to re-run the checks) and see if that fixes it. Elsewise I'll look into it tomorrow.

mvanbeusekom (approved):
LGTM

PR #897, opened 2021-11-19
Removed the permission API query in the geolocator_web package and updated the README.md

florissmit1:
Removed the Permission API query in the geolocator_web package, to be found in the `checkPermission` method.

Currently, apps running in browsers that don't support the JavaScript Permissions API, will stop working when using the `checkPermission` method.

The `checkPermission` method will always return the `LocationPermission.denied` status.

No

Does not apply.

Does not apply

### DIFF
--- a/geolocator_web/lib/geolocator_web.dart
+++ b/geolocator_web/lib/geolocator_web.dart
@@ -15,7 +15,6 @@ export 'package:geolocator_platform_interface/geolocator_platform_interface.dart
 ///
 /// This class implements the `package:geolocator` functionality for the web.
 class GeolocatorPlugin extends GeolocatorPlatform {
-  static const _permissionQuery = {'name': 'geolocation'};
 
   final GeolocationManager _geolocation;
   final PermissionsManager _permissions;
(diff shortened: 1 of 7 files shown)


geolocator_web/lib/geolocator_web.dart
```diff
@@ -55,9 +53,7 @@ class GeolocatorPlugin extends GeolocatorPlatform {
       );
     }
 
-    return await _permissions.query(
-      _permissionQuery,
-    );
+    return LocationPermission.denied;
```
mvanbeusekom, line 56: We should only do this for browsers that don't support the `query` method. For browser that do support the `query` method it should still be available. In this case the best way to resolve the issue would be in the `HtmlPermissionManager` class and update the `query` method to:

```dart
Future<LocationPermission> query() async {
    final permissions = _permissions; 
    if (permissions == null) {
      return LocationPermission.denied;
    }
  
    final status = await permissions.query(
      _permissionQuery,
    );
      
    return status.state != null 
      ? toLocationPermission(status.state) 
      : LocationPermission.denied;
}
```
This method (and the whole `GeolocatorWeb` class) doesn't require any changes.

I have tested this (a slightly different version) using DartPad on several browsers and it works.

mvanbeusekom (changes requested):

PR #933, opened 2022-01-06
Export `ActivityType` at `geolocator.dart`

mduccc:
enhancement

When I use `AppleSettings` with `activityType` property. `AppleSettings` has exported, but the type of the property itself is not.
 Must import: 
`import 'package:geolocator_apple/geolocator_apple.dart';`

On the opposite side, unnecessary below import for `AndroidSettings`:
`import 'package:geolocator_android/geolocator_android.dart';`

Export `ActivityType` at `geolocator.dart`

Then all need to do:
`import 'package:geolocator/geolocator.dart';`

### DIFF
--- a/geolocator/lib/geolocator.dart
+++ b/geolocator/lib/geolocator.dart
@@ -8,7 +8,8 @@ import 'package:geolocator_platform_interface/geolocator_platform_interface.dart
 export 'package:geolocator_platform_interface/geolocator_platform_interface.dart';
 export 'package:geolocator_android/geolocator_android.dart'
     show AndroidSettings;
-export 'package:geolocator_apple/geolocator_apple.dart' show AppleSettings;
+export 'package:geolocator_apple/geolocator_apple.dart'
+    show AppleSettings, ActivityType;
 
 /// Wraps CLLocationManager (on iOS) and FusedLocationProviderClient or
 /// LocationManager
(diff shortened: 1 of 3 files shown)


geolocator/lib/geolocator.dart
```diff
@@ -8,7 +8,7 @@ import 'package:geolocator_platform_interface/geolocator_platform_interface.dart
 export 'package:geolocator_platform_interface/geolocator_platform_interface.dart';
 export 'package:geolocator_android/geolocator_android.dart'
     show AndroidSettings;
-export 'package:geolocator_apple/geolocator_apple.dart' show AppleSettings;
+export 'package:geolocator_apple/geolocator_apple.dart';
```
mvanbeusekom, line 11: The `ActivityType` should indeed be exported. However it would be better to add it to the `show` condition. 

We are in the progress of adding more logic to the `geolocator_apple` package and we don't want to expose everything by default through the `geolocator` app-facing package.

```suggestion
export 'package:geolocator_apple/geolocator_apple.dart' 
         show AppleSettings, ActivityType;
```

mvanbeusekom (changes requested):
Thank you for this submission and apologies for the late reply. I left one remark and would gladly merge your PR if you could provide a fix.

geolocator/lib/geolocator.dart
mduccc: > 

Yes. I fixed it!

mvanbeusekom (approved):
LGTM, thank you very much for this contribution.

PR #958, opened 2022-01-24
Update README.md

pinguluk:
- Position? can be null
- Used interpolation to compose strings and values

Docs update


mvanbeusekom (approved):
LGTM, thanks for the update.

PR #978, opened 2022-02-11
Adds Android foreground location support

Wackymax:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)

Feature: Adds support for background (Foreground service) support to Android for when the applications goes into the background while location stream is active

Related to PR: https://github.com/Baseflow/flutter-geolocator/pull/957

### :arrow_heading_down: What is the current behavior?

Currently when the application goes into background and the OS kills the activity then location streaming stops.

### :new: What is the new behavior (if this is a feature change)?

New options have been introduced on Android that will allow the users to configure the background behaviour allow the service to continue running in the background keeping the activity alive while the location stream is active.

### :boom: Does this PR introduce a breaking change?

A new Android service is introduced which changes the behaviour of the location stream to use the new service but existing apps will not know about this change.

### :bug: Recommendations for testing

Testing the normal mode to validate the location stream is still working as expected and testing the background mode with the application running in the background.

### :memo: Links to relevant issues/docs

https://github.com/Baseflow/flutter-geolocator/issues/948
https://github.com/Baseflow/flutter-geolocator/issues/53

### :thinking: Checklist before submitting

- [X] I made sure all projects build.
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning).
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I followed the style guide lines ([code style guide](https://github.com/Baseflow/flutter-geolocator/blob/master/CONTRIBUTING.md)).
- [X] I updated the relevant documentation.
- [X] I rebased onto current `master`.

Issue #948, opened 2022-01-14
iOS Background Location


Wackymax:
## 🚀 Feature Requests

Hi, would it be possible to add a way for us to tweak the background location setting of the iOS location manager?

### Contextualize the feature

```
AppleSettings(
        accuracy: LocationAccuracy.high,
        activityType: ActivityType.airborne,
        allowsBackgroundLocationUpdates: true,
        showsBackgroundLocationIndicator: true,
      )
```

Would set

```
if (@available(iOS 9.0, *)) {
                self.clLocationManager.allowsBackgroundLocationUpdates = enable;
            }
            if (@available(iOS 11.0, *)) {
                self.clLocationManager.showsBackgroundLocationIndicator = enable;
            }
```

### Describe the feature

Adding addition options for Apple Settings to configure the clLocationManager allowsBackgroundLocationUpdates and showsBackgroundLocationIndicator.

### Platforms affected (mark all that apply)
- [X] :iphone: iOS
- [ ] :robot: Android

mvanbeusekom:
Hi @Wackymax, 

There is a way to do so but is not documented. You can enable the `allowsBackgroundLocationUpdates` property by adding the `location` key to the `UIBackGrounds` collection in your `ios/Runner/Info.plist` file. The easiest way to do this is by opening your project in Xcode (make sure you open the `ios` folder in your project) and add the Background Modes capability to your XCode project (`Project > Signing and Capabilities > "+ Capability"` button) and select "Location Updates".

The current implementation looks like this ([GeolocationHandler.m#L117-123(https://github.com/Baseflow/flutter-geolocator/blob/291ebf052ac562756fae67242d1db45114031a1c/geolocator_apple/apple/Classes/Handlers/GeolocationHandler.m#L117)]:

```objc
+ (BOOL) shouldEnableBackgroundLocationUpdates {
    if (@available(iOS 9.0, *)) {
        return [[NSBundle.mainBundle objectForInfoDictionaryKey:@"UIBackgroundModes"] containsObject: @"location"];
    } else {
        return NO;
    }
}
```

and is called on [line 71](https://github.com/Baseflow/flutter-geolocator/blob/291ebf052ac562756fae67242d1db45114031a1c/geolocator_apple/apple/Classes/Handlers/GeolocationHandler.m#L71)when you start listing to the stream returned from `getPositionStream()`:

```objc
#if TARGET_OS_IOS
[...]

Issue #53, opened 2018-07-31
Add support for background location tracking


martijn00:
Also see: https://github.com/flutter/plugins/pull/656

eliashdezr:
This is one of the main reasons I haven't switch one of my apps to Flutter, background location tracking is a must.

aidandavis:
I'm also looking into this. Wanting to build an app in a corporate setting with a background GPS tracking requirement. flutter/plugins#656 (mentioned above) isn't on Pub yet and only supports iOS.
I'll subscribe to this thread for any updates.

aidandavis:
I ended up writing my own implementation, based off [this Google sample](https://github.com/googlesamples/android-play-location/tree/master/LocationUpdatesForegroundService) that runs when the app closes.

I'm pushing the location data to firebase, so I didn't need to send anything back up the method channel. I just upload straight to Firebase using native code when a location result arrives. This means I can't generalise it to help this plugin out, but that sample repo might be useful.

velegip:
I found a solution to generalize your @aidandavis method.
In this [packege](https://github.com/talkdirty/background_location_updates) all locations stored in sqlite.
If your foreground service can accept store provider sqlite/firebase/... .
With MethodChannel we can get the db file location, and the flutter app can read it.

aidandavis:
That's definitely a way to do it. The problem still remaining is doing something with the data while the app is in background or the sytem chooses to close it.

With regards to Doze, it should only apply when the phone is stationary (like on a table). There's no point tracking location updates at that point anyway - the last location sent prior to Doze would still be accurate.

sestegra:
On top of this, Dart code could be embedded as the same way implemented in alarm_manager_android. The Android callback calls Dart code without the app running.

dark-chocolate:
Guys, any update on this? This feature can add wings to this plugin. It is must needed.

ibrierley:
It's maybe also worth noting that some devices can't get gps locations with the screen off. I've been stuck on this one for a while, and finally found it's down to the devices chipset. So it may be worth double checking that as well, especially if it's a low battery device like a watch.

dark-chocolate:
@ibrierley I am not sure about the wearables but all the devices running API >= 16 are capable of getting location updates even when the screen is off. I tested that myself too.

ibrierley:
[...]

PR #980, opened 2022-02-11
Add foreground service location support

Wackymax:
Feature: Adds support for background (Foreground service) support to Android for when the applications goes into the background while location stream is active.

This is dependent on the following PR being released:
https://github.com/Baseflow/flutter-geolocator/pull/978

Currently when the application goes into background and the OS kills the activity then location streaming stops.

New options have been introduced on Android that will allow the users to configure the background behaviour allow the service to continue running in the background keeping the activity alive while the location stream is active.

A new Android service is introduced which changes the behaviour of the location stream to use the new service but existing apps will not know about this change.

Testing the normal mode to validate the location stream is still working as expected and testing the background mode with the application running in the background.

https://github.com/Baseflow/flutter-geolocator/issues/948
https://github.com/Baseflow/flutter-geolocator/issues/53


geolocator/pubspec.yaml
```diff
@@ -26,8 +26,8 @@ dependencies:
   flutter:
     sdk: flutter
   
-  geolocator_platform_interface: ^4.0.0
-  geolocator_android: ^3.0.2
+  geolocator_platform_interface: ^4.0.0+1
```
mvanbeusekom, line 29: Version 4.0.0+1 doesn't exists, this should probably be updated to version 4.0.3.
```suggestion
  geolocator_platform_interface: ^4.0.3
```

mvanbeusekom (changes requested):
To be able to publish this update we should:

[ ] Update the version number in the `pubspec.yaml`;
[ ] Add release note the the `CHANGELOG.md`;
[ ] Update the example App to include new changes;
[ ] Add documentation on how to use the new background services in the `README.md` file.

P.S. I would also like to add the iOS changes you made. They seemed really useful. If you don't have the time, do you mind if I copy it and create a PR my self?

Wackymax:
> To be able to publish this update we should:
> 
> * [x]  Update the version number in the `pubspec.yaml`;
> * [x]  Add release note the the `CHANGELOG.md`;
> * [ ]  Update the example App to include new changes;
> * [ ]  Add documentation on how to use the new background services in the `README.md` file.
> 
> P.S. I would also like to add the iOS changes you made. They seemed really useful. If you don't have the time, do you mind if I copy it and create a PR my self?

The changes have been applied. I only updated the readme, the example app does not set any settings for Apple or Android at the moment. In terms of the iOS changes, I don't mind bringing them in. I would be happy to help out there if you feel like it will add value to the library.

mvanbeusekom:
That would be great. I would like to merge the iOS changes first (in a separate PR) and then update this PR to include the updated geolocator_Apple package as well. 

That way we can release everything in one go.

Wackymax:
> That would be great. I would like to merge the iOS changes first (in a separate PR) and then update this PR to include the updated geolocator_Apple package as well.
> 
> That way we can release everything in one go.

[982](https://github.com/Baseflow/flutter-geolocator/pull/982) has been created with the iOS changes.

mvanbeusekom:
@Wackymax just released version 2.1.1 of the geolocator_apple package (thanks for that). 

Could you update this PR to reference this new version of the package and maybe add a little bit of documentation regarding the new parameter to the README.md file?

Wackymax:
> @Wackymax just released version 2.1.1 of the geolocator_apple package (thanks for that).
> 
> Could you update this PR to reference this new version of the package and maybe add a little bit of documentation regarding the new parameter to the README.md file?

This branch has been updated with the Apple settings changes.

Wackymax:
@mvanbeusekom it seems like something in the iOS 2.1.1 release is broken. I am not entirely sure what or how to fix it though  🤔

mvanbeusekom:
I will have a look in an hour or so when I am behind my desk.

Wackymax:
I have not confirmed this with the published package but it seems like the published iOS may be breaking the official geolocator package as well because we aren't pinning versions in this package. Maybe we should remove the ^ on the platform packages to prevent them from getting upgraded automatically?

mvanbeusekom:
> I have not confirmed this with the published package but it seems like the published iOS may be breaking the official geolocator package as well because we aren't pinning versions in this package. Maybe we should remove the ^ on the platform packages to prevent them from getting upgraded automatically?

There is indeed a strange issue, I am currently investigating more detailed.

We should not remove the `^` symbol as in many cases we would like to silently release fixes to the platform specific packages without having users to update the `geolocator` plugin it self.

mvanbeusekom:
So something strange is happening. Somehow the `.h` and `.m` files in the `ios` and `macos` folders (which are all symbolic links to the matching file in the `apple` folder lost their link attribute. 

This means that now they are just files containing a string similar to this:
```
../../../apple/Classes/Handlers/PermissionHandler.m
```

This is of course resulting in a compile error as this is not valid Objective-C code.

mvanbeusekom:
(sorry, closed the PR by accident)

mvanbeusekom:
I think I know what happend, I have published the `geolocator_apple` package from a Windows machine (as that is what I had close at hand yesterday). 

I didn't realise this would delete all symbolic links when publishing to pub.dev. I will publish a hotfix now using macOS machine which should fix the issue. 

Note to self: never publish Apple stuff from a Windows machine.

mvanbeusekom (approved):
LGTM, thanks again for all the work @Wackymax, really appreciate it.

I will merge and release the update today. Do you mind if I mention you in a tweet announcing the update (and do you have a twitter handle)?

Wackymax:
Wow that is such a bizarre issue 😅. Glad you found it!

Thanks that is pretty awesome! It was really fun contributing, will definitely contribute again when I see use case. My twitter handle is @HennieBrink3

PR #998, opened 2022-03-04
Clear stream before emitting error

zbarbuto:

Bug fix

When the location stream times out, if your handler for this error is to immediately re-try listening to the stream, you receive a dead stream.

Clear out the stale stream before throwing the error so re-listening will get a fresh stream.

No

Something along these lines (but force it to timeout)

```dart
void _setupLocationStream() {
  final stream = Geolocator.getPositionStream(
    locationSettings: LocationSettings(timeLimit: Duration(seconds: 3))
  );
  locationSubscription = stream.listen((location) {
    print(location);
  });
   locationSubscription!.onError((error) {
     // This attempt to re-setup the stream will get the already timed out stream, meaning
     // you do not get any more updates.
     // If you await a Future.delayed(Duration(milliseconds: 100)); here, you do get updates
     _setupLocationStream();
   });
}
```

Specifically [this check](https://github.com/NextFaze/flutter-geolocator/blob/master/geolocator_platform_interface/lib/src/implementations/method_channel_geolocator.dart#L156-L158) is not null because the call to check if the existing stream is null fails (verified with breakpoints)

### DIFF
--- a/geolocator_platform_interface/CHANGELOG.md
+++ b/geolocator_platform_interface/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.0.4
+
+- Fixes a bug where listening to the position stream immediately after an error, results in listening to a dead stream. 
 
 ## 4.0.3
 
--- a/geolocator_platform_interface/lib/src/implementations/method_channel_geolocator.dart
+++ b/geolocator_platform_interface/lib/src/implementations/method_channel_geolocator.dart
@@ -167,12 +167,12 @@ class MethodChannelGeolocator extends GeolocatorPlatform {
       positionStream = positionStream.timeout(
         timeLimit,
         onTimeout: (s) {
+          _positionStream = null;
           s.addError(TimeoutException(
             'Time limit reached while waiting for position update.',
             timeLimit,
           ));
           s.close();
-          _positionStream = null;
         },
       );
     }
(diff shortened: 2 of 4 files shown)


geolocator_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.0.4
+
+- Fix listening location to stream immediately on error
```
mvanbeusekom, line 3: We try to write change log entries starting with a third person singular verb. How about something like:

```suggestion
- Fixes a bug where listening to the position stream immediately after an error, results in listening to a dead stream. 
```

mvanbeusekom (changes requested):
Hi @zbarbuto,

Thank you for this contribution, good catch. 

Could you also provide a unit test that proves the previous situation is broken and this fix resolves the issue? That will help us to prevent the same bug in the future.

geolocator_platform_interface/CHANGELOG.md
zbarbuto: Done, thanks

zbarbuto:
Have added a test and squashed commits as cleanup.

Sorry the test is a bit clumsy, was a bit complex to reproduce the issue without the fix in testing environment.

mvanbeusekom (approved):
LGTM, thanks again for this contribution

Issue #1012, opened 2022-03-17
Geolocator with foreground task package - "location background service has not started correctly"


cheeco83:
I try to run the Geolocator with the "Foreground Task" package to retrieve current position. Even though I am able to get the Positions using getCurrentPosition(), setting up a Stream fails with the error message "Location background service has not started correctly". I do not know what went wrong here, but the combination of Geolocator and Foreground Task was even given as an example and thus supposed to work earlier.

Interestingly, apparently nobody else has this particular problem, as I get no hits in Google when looking for this error. I also do not know in which package the problem is located. I prepared a minimalistic example that reproduces the error in the annex. Could you point me in the right direction where to look for?

Full disclosure: I already posted a similar question in the "Foreground Task" issue tracker: https://github.com/Dev-hwang/flutter_foreground_task/issues/66

Thank you!

Here is the example: [main.dart.txt](https://github.com/Baseflow/flutter-geolocator/files/8292765/main.dart.txt)

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

panalgin:
StreamHandlerImpl.java line ~87
```java
    if (foregroundLocationService == null) {
      Log.e(TAG, "Location background service has not started correctly");
      return;
    }
```

This line simply breaks an external foreground service to listen geolocator updates. I think this came with @Wackymax latest additions to run the geolocator package with internal foreground service he implemented.

I think it prevents so many customizations one would do using an external foreground service packages like https://github.com/Dev-hwang/flutter_foreground_task/ and https://pub.dev/packages/flutter_background_service

Also, speaking of battery-wise is it a good idea to have GPS chip be turned on always? This is what geolocator is doing it seems, if I use the ForegroundNotificationConfig and start the service in foreground mode (or background technical confusions both here and in the source code)


I was thinking of using a foreground service that would check the GPS location once per 5 minutes and if it is like 50 meters away from the last known coordinates, then start a StreamSubscription and listen to the changes until the client becomes idle/still + exponential backoff. Or maybe a trigger mechanism to start listening to one's GPS coordinates once he/she starts walking, with the help of ACTIVITY_RECOGNITION API perhaps, anyone has any idea of the feasibility?

panalgin:
Temporary fix:  use geolocator: 8.0.1  in pubspec.yaml and you can listen to position change stream

cheeco83:
[...]

PR #1016, opened 2022-03-26
Remove activity dependency

Wackymax:

Bug fix

Currently when someone tries to obtain a position update stream from a different flutter engine that isn't attached to an activity an error is returned and the location updates aren't being returned.

This PR introduces changes that allows the geolocator foreground service to bind to the application context instead of an activity which will allow other flutter background libraries to start the service from the background in absence of an Android activity. This also now determine wether to to use our own foreground service or to just start the location updates in the service asking for the position stream.

No

The position stream will now behave differently depending on whether foreground notification config has been provided. If not then a position stream is created like it used to before the foreground service was introduced, if it is provided the the foreground service is used instead for the position updates. I've introduced log messages to indicate this distinction. Other libraries like `flutter_foreground_task` should also work correctly now where a new flutter engine is created and bound to this plugin.

https://github.com/Baseflow/flutter-geolocator/issues/1012

### DIFF
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java
@@ -91,8 +91,8 @@ public class GeolocatorPlugin implements FlutterPlugin, ActivityAware {
 
     LocationServiceHandlerImpl locationServiceHandler = new LocationServiceHandlerImpl();
     locationServiceHandler.startListening(registrar.context(), registrar.messenger());
-    locationServiceHandler.setContext(registrar.activity());
-    geolocatorPlugin.bindForegroundService(registrar.activity());
+    locationServiceHandler.setContext(registrar.activeContext());
+    geolocatorPlugin.bindForegroundService(registrar.activeContext());
   }
 
   @Override
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java
@@ -8,6 +8,7 @@ import android.os.IBinder;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+
 import com.baseflow.geolocator.location.GeolocationManager;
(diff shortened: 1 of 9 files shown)


cheeco83:
Wackymax, I tried to implement your changes. Unfortunately, I get the same "Foreground task has not started correctly" is still there.

But I might have made a mistake getting your updated github commit. In the Pubspec.yaml, I added the following lines and ran pub upgrade geolocator afterwards. I am not sure if I have the right commit hash, as there were three this day.

  geolocator:
    git:
      url: https://github.com/Wackymax/flutter-geolocator.git
      ref: 06379fd  # commit hash
      path: geolocator/

Wackymax:
@mvanbeusekom can you kick off the validations again? Looks like something went wrong during the build process.

Wackymax:
Done 😊

geolocator_android/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java
```diff
@@ -73,7 +91,8 @@ public static void registerWith(io.flutter.plugin.common.PluginRegistry.Registra
 
     LocationServiceHandlerImpl locationServiceHandler = new LocationServiceHandlerImpl();
     locationServiceHandler.startListening(registrar.context(), registrar.messenger());
-    locationServiceHandler.setActivity(registrar.activity());
+    locationServiceHandler.setContext(registrar.activity());
+    geolocatorPlugin.bindForegroundService(registrar.activity());
```
mvanbeusekom, line 95: I am aware this is all deprecated but non the less it might be better to use the `registrar.activeContext()` in this case. The `activeContext()` method will return the `activity` if it's available and otherwise will return the Application's context (official documentation can be found [here](https://api.flutter.dev/javadoc/io/flutter/plugin/common/PluginRegistry.Registrar.html#activeContext())).
```suggestion
    locationServiceHandler.setContext(registrar.activeContext());
    geolocatorPlugin.bindForegroundService(registrar.activeContext());
```

mvanbeusekom (changes requested):
Hi @Wackymax,

Thank you for this pull request, looks very good. I just have two small remarks, would appreciate it if you could have a look.

mvanbeusekom (approved):
Thank you for the update! LGTM

PR #1101, opened 2022-07-12
Removing redundant code

nikb7:
Removing redundant code to check permissions and get location
isLocationServiceEnabled method is not required here.
If the device location is off, in that case when we call getCurrentPosition() method, it requests for location from the device. And the pop up to turn on location comes.

### DIFF
--- a/geolocator/README.md
+++ b/geolocator/README.md
@@ -165,18 +165,8 @@ import 'package:geolocator/geolocator.dart';
 /// When the location services are not enabled or permissions
 /// are denied the `Future` will return an error.
 Future<Position> _determinePosition() async {
-  bool serviceEnabled;
   LocationPermission permission;
-
-  // Test if location services are enabled.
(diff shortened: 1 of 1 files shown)


mvanbeusekom:
Unfortunately this is only the case on Android, all other platforms don't support this behavior and the current code is a valid way to validate if the services are enabled or not.

nikb7:
@mvanbeusekom Wouldn't it better to add a check for platform in the code instead of doing this?
Since GeoLocator doesn't have any method which can request device for enabling location. Hence most user will just end up using the openLocationSetting method, which is not the best way to handle this. Specially for android!

nikb7:
> @mvanbeusekom Wouldn't it better to add a check for platform in the code instead of doing this? Since GeoLocator doesn't have any method which can request device for enabling location. Hence most user will just end up using the openLocationSetting method, which is not the best way to handle this. Specially for android!

We should at least add a comment about this behaviour on android

PR #1108, opened 2022-07-19
Fix Nullpointer in getLocationAccuracy on Android

### LINKED ISSUE #899: Sometimes in logs we can see NPE
Our testers reported that sometimes in logs they can see npe. Looks like it happens when we trying to get location first time after user give us all required permissions:
 
```
MethodChannel#flutter.baseflow.com/geolocator: Failed to handle method call
MethodChannel#flutter.baseflow.com/geolocator: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Enum.ordinal()' on a null object reference
MethodChannel#flutter.baseflow.com/geolocator:         at c.b.a.l.a(Unknown Source:13)
MethodChannel#flutter.baseflow.com/geolocator:         at c.b.a.l.G(Unknown Source:111)
MethodChannel#flutter.baseflow.com/geolocator:         at e.a.c.a.j$a.a(Unknown Source:17)
MethodChannel#flutter.baseflow.com/geolocator:         at io.flutter.embedding.engine.e.b.e(Unknown Source:57)
MethodChannel#flutter.baseflow.com/geolocator:         at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(Unknown Source:4)
MethodChannel#flutter.baseflow.com/geolocator:         at android.os.MessageQueue.nativePollOnce(Native Method)
MethodChannel#flutter.baseflow.com/geolocator:         at android.os.MessageQueue.next(MessageQueue.java:335)
MethodChannel#flutter.baseflow.com/geolocator:         at android.os.Looper.loop(Looper.java:206)
MethodChannel#flutter.baseflow.com/geolocator:         at android.app.ActivityThread.main(ActivityThread.java:8595)
MethodChannel#flutter.baseflow.com/geolocator:         at java.lang.reflect.Method.invoke(Native Method)
MethodChannel#flutter.baseflow.com/geolocator:         at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:602)
MethodChannel#flutter.baseflow.com/geolocator:         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1130)
```

kuhnroyal:
Fixes #899

### DIFF
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
@@ -158,7 +158,7 @@ class MethodCallHandlerImpl implements MethodChannel.MethodCallHandler {
             context,
             (ErrorCodes errorCode) ->
                 result.error(errorCode.toString(), errorCode.toDescription(), null));
-    if (result != null) {
+    if (status != null) {
       result.success(status.ordinal());
     }
   }
--- a/geolocator_android/CHANGELOG.md
+++ b/geolocator_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.2
+
+- Fix nullpointer in `MethodCallHandlerImpl.getLocationAccuracy`
+
 ## 4.0.1
 
 - Migrates to Dart SDK 2.15.0 and Flutter 2.8.0.
--- a/geolocator_android/pubspec.yaml
+++ b/geolocator_android/pubspec.yaml
@@ -2,7 +2,7 @@ name: geolocator_android
 description: Geolocation plugin for Flutter. This plugin provides the Android implementation for the geolocator.
 repository: https://github.com/baseflow/flutter-geolocator/tree/main/geolocator_android
 issue_tracker: https://github.com/baseflow/flutter-geolocator/issues?q=is%3Aissue+is%3Aopen
-version: 4.0.1
+version: 4.0.2
 
 environment:
   sdk: ">=2.15.0 <3.0.0"


geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
```diff
@@ -158,7 +158,7 @@ private void getLocationAccuracy(MethodChannel.Result result, Context context) {
             context,
             (ErrorCodes errorCode) ->
                 result.error(errorCode.toString(), errorCode.toDescription(), null));
-    if (result != null) {
+    if (status != null) {
```
mvanbeusekom, line 161: The check should be on both parameters as `result` can also be `null`:

```suggestion
    if (status != null && result != null) {
```

mvanbeusekom (changes requested):
Thanks for the contribution, it contains one small error though.

geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
kuhnroyal: Yea I had that at first, but this is not being checked anywhere else either.
If `result` can be null, then this will error everywhere in the file (see line above).

And also I don't think it can be null because of line 56:
```java
public void onMethodCall(@NonNull MethodCall call, @NonNull MethodChannel.Result result) {
```

mvanbeusekom:
@kuhnroyal thank you for the fix. I have just released version 4.0.2 of the [geolocator_android](https://pub.dev/packages/geolocator_android) package which should become available locally after running `flutter pub upgrade`.

PR #1121, opened 2022-08-16
Prepare for fix to https://github.com/flutter/flutter/issues/109339.

stereotype441:
Currently, in some circumstances where a subclasses of
`PlatformInterface` erroneously uses `implements` rather than
`extends`, a `NoSuchMethodError` will be thrown (in spite of the
documentation at
https://pub.dev/documentation/plugin_platform_interface/latest/plugin_platform_interface/PlatformInterface/verify.html
claiming that `AssertionError` will be thrown).

After https://github.com/flutter/flutter/issues/109339 is fixed, the
correct type of exception will be thrown.  To avoid a test breakage in
`geolocator_platform_interface_test.dart` when the fix happens, we
need to modify the test so that it doesn't care what kind of exception
is thrown.

Bug fix

The test expects `NoSuchMethodError` to be thrown; this will break after https://github.com/flutter/flutter/issues/109339 is fixed.

The test doesn't care what kind of exception is thrown, so fixing https://github.com/flutter/flutter/issues/109339 will not cause a breakage.

No.

### DIFF
--- a/geolocator_platform_interface/test/geolocator_platform_interface_test.dart
+++ b/geolocator_platform_interface/test/geolocator_platform_interface_test.dart
@@ -19,7 +19,14 @@ void main() {
     test('Cannot be implemented with `implements`', () {
       expect(() {
         GeolocatorPlatform.instance = ImplementsGeolocatorPlatform();
-      }, throwsNoSuchMethodError);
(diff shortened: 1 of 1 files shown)


mvanbeusekom (approved):
LGTM, thank you for the heads-up very much appreciated.

Issue #1142, opened 2022-10-01
Android foreground notification is not cleared after position Stream is cancelled


ValentinTaleb:
## 🐛 Bug Report

On Android, it's possible display a foreground notification in order to keep recording location updates even after the user has moved the app in the background.

This notification does not disappear after the position stream as been cancelled.

### Expected behavior

The Android foreground notification should disappear once the position stream has been cancelled.

### Reproduction steps

```
StreamSubscription<Position>? _positionStreamSubscription;

@override
void initState() {
  super.initState();

  if (Platform.isAndroid) {
    locationSettings = AndroidSettings(
      ...
      foregroundNotificationConfig: ForegroundNotificationConfig(
        notificationTitle: 'Notification title',
        notificationText: 'Notification text',
        notificationIcon: const AndroidResource(name: 'ic_notification', defType: 'drawable'),
      ));
  }

  _positionStreamSubscription = Geolocator.getPositionStream(locationSettings: locationSettings).listen(
    ...
  );
}

@override
void dispose() {
  _positionStreamSubscription.cancel();

  super.dispose();
}
```

### Configuration

**Version:** 9.0.2

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

Thank you very much in advance

ValentinTaleb:
Am I the only one who experiences this issue?

Juan-EZ:
I am experiencing the same issue. The notification persists even after I force close the app on Android

fleoparra:
Same problem here. Notification continues after cancel the stream.

MrCheese02:
I can confirm the same problem. After all subscriptions are cancelled I receive the following log messages:
```
E/FlutterGeolocator( 9409): Geolocator position updates stopped
E/FlutterGeolocator( 9409): There is still another flutter engine connected, not stopping location service
E/FlutterGeolocator( 9409): Geolocator position updates started
```
However, upon the app's and the Geolocator service's startup it only reports about one connection to the Flutter engine.
```
D/FlutterGeolocator(11402): Attaching Geolocator to activity
D/FlutterGeolocator(11402): Creating service.
D/FlutterGeolocator(11402): Binding to location service.
D/FlutterGeolocator(11402): Geolocator foreground service connected
D/FlutterGeolocator(11402): Initializing Geolocator services
D/FlutterGeolocator(11402): Flutter engine connected. Connected engine count 1
```

habanoz:
[...]

Issue #986, opened 2022-02-19
Leak on terminated app


bujdy:
## 🐛 Bug Report

I wanted to try to listen to locations even after i terminate app so i added 
```foregroundNotificationConfig: const ForegroundNotificationConfig(
            notificationText:
            "Example app will continue to receive your location even when you aren't using it",
            notificationTitle: "Running in Background",
            enableWakeLock: true,
          )
```
to settings. But when i terminate the app i get this leak 

```
E/ActivityThread(13871): Activity com.enviroapp.enviroapp.MainActivity has leaked ServiceConnection com.baseflow.geolocator.GeolocatorPlugin$1@576aec9 that was originally bound here
E/ActivityThread(13871): android.app.ServiceConnectionLeaked: Activity com.enviroapp.enviroapp.MainActivity has leaked ServiceConnection com.baseflow.geolocator.GeolocatorPlugin$1@576aec9 that was originally bound here
E/ActivityThread(13871): 	at android.app.LoadedApk$ServiceDispatcher.<init>(LoadedApk.java:1811)
E/ActivityThread(13871): 	at android.app.LoadedApk.getServiceDispatcherCommon(LoadedApk.java:1683)
E/ActivityThread(13871): 	at android.app.LoadedApk.getServiceDispatcher(LoadedApk.java:1662)
E/ActivityThread(13871): 	at android.app.ContextImpl.bindServiceCommon(ContextImpl.java:1819)
E/ActivityThread(13871): 	at android.app.ContextImpl.bindService(ContextImpl.java:1749)
E/ActivityThread(13871): 	at android.content.ContextWrapper.bindService(ContextWrapper.java:756)
E/ActivityThread(13871): 	at com.baseflow.geolocator.GeolocatorPlugin.onAttachedToActivity(GeolocatorPlugin.java:126)
E/ActivityThread(13871): 	at io.flutter.embedding.engine.FlutterEngineConnectionRegistry.add(FlutterEngineConnectionRegistry.java:155)
E/ActivityThread(13871): 	at io.flutter.plugins.GeneratedPluginRegistrant.registerWith(GeneratedPluginRegistrant.java:29)
E/ActivityThread(13871): 	at java.lang.reflect.Method.invoke(Native Method)
E/ActivityThread(13871): 	at io.flutter.embedding.engine.plugins.util.GeneratedPluginRegister.registerGeneratedPlugins(GeneratedPluginRegister.java:80)
[...]

PR #1153, opened 2022-10-14
Android location stream not cancelling

Wackymax:

Bug fix

Currently when the location stream is cancelled the geolocator background service does not fully shutdown correctly.

This fixes the check for when multiple flutter engines are connected and location stream is explicitly being cancelled by the user instead of being cancelled by the activity process getting destroyed.

No

1. Set the Android foreground config. Start the location stream and shortly after stop it. The foreground notification should be removed when the service shuts down.
2. When using a background service like flutter_background_service and starting the geolocator stream in the background service and then killing the main activity should not stop the geolocator service and it should continue to send position updates to the background service

https://github.com/Baseflow/flutter-geolocator/issues/1142
https://github.com/Baseflow/flutter-geolocator/issues/986

### DIFF
--- a/geolocator_android/CHANGELOG.md
+++ b/geolocator_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.4
+
+- Fix a bug where the location service would not stop correctly when the location stream is cancelled.
+
 ## 4.1.3
 
 - Export `AndroidResource` class at `geolocator_android.at`.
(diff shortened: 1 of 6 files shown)


Wackymax:
@mvanbeusekom I didn't make changes to the dart code so not sure what the codecov is complaining about?

mvanbeusekom (approved):
LGTM, thanks for fixing this @Wackymax

PR #1155, opened 2022-10-20
Implement allowBackgroundLocationUpdates iOS setting

landyrev:
Feature

Currently geolocator sets `allowBackgroundLocationUpdates` equal to `true` by default when the `UIBackgroundModes` contains `location` value inside the app's Info.plist, which leads to `Geolocator.getPositionStream` receiving background geolocations even if you don't need them

User can disable background geolocations updates even if their app contains `location` value inside the `UIBackgroundModes` key

No, `allowBackgroundLocationUpdates` is set to true by default

Can be tested on iOS devices and simulators
1. Add `location` value to the `UIBackgroundModes` Info.plist key
2. Request position stream with `getPositionStream` with `locationSettings` equal to:
```dart
AppleSettings(
  allowBackgroundLocationUpdates: false
)
```
3. Send app to the background 
4. Check that there is no background geolocation indicator visible in status bar

### DIFF
--- a/geolocator_apple/ios/Classes/Handlers/GeolocationHandler.m
+++ b/geolocator_apple/ios/Classes/Handlers/GeolocationHandler.m
@@ -65,7 +65,8 @@ double const kMaxLocationLifeTimeInSeconds = 5.0;
                pauseLocationUpdatesAutomatically:NO
                                     activityType:CLActivityTypeOther
                    isListeningForPositionUpdates:self.isListeningForPositionUpdates
-                 showBackgroundLocationIndicator:NO];
+                 showBackgroundLocationIndicator:NO
+                  allowBackgroundLocationUpdates:NO];
 }
 
 - (void)startListeningWithDesiredAccuracy:(CLLocationAccuracy)desiredAccuracy
@@ -73,6 +74,7 @@ double const kMaxLocationLifeTimeInSeconds = 5.0;
         pauseLocationUpdatesAutomatically:(BOOL)pauseLocationUpdatesAutomatically
           showBackgroundLocationIndicator:(BOOL)showBackgroundLocationIndicator
                              activityType:(CLActivityType)activityType
+           allowBackgroundLocationUpdates:(BOOL)allowBackgroundLocationUpdates
                             resultHandler:(GeolocatorResult _Nonnull )resultHandler
                              errorHandler:(GeolocatorError _Nonnull)errorHandler {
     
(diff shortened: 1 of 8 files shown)


geolocator_apple/ios/Classes/Handlers/GeolocationHandler.m
```diff
@@ -105,7 +109,8 @@ - (void)startUpdatingLocationWithDesiredAccuracy:(CLLocationAccuracy)desiredAccu
   
 #if TARGET_OS_IOS
   if (@available(iOS 9.0, macOS 11.0, *)) {
-    locationManager.allowsBackgroundLocationUpdates = [GeolocationHandler shouldEnableBackgroundLocationUpdates];
+    locationManager.allowsBackgroundLocationUpdates = allowBackgroundLocationUpdates
+      && [GeolocationHandler shouldEnableBackgroundLocationUpdates];
```
mvanbeusekom, line 113: There should be a unit test to verify the `allowsBackgroundLocationUpdates` property is set correctly (only when both values are true.

mvanbeusekom (changes requested):
Thank you for contributing, this looks very good. I have just one request to add an additional test.

geolocator_apple/ios/Classes/Handlers/GeolocationHandler.m
landyrev: @mvanbeusekom good point, thanks! Added a couple of tests for that

mvanbeusekom (approved):
LGTM, thanks for this great addition.

Faiizii:
I think it should be in readme file. I implemented the background location updates in one of my app. In the recent updates it stopped working with the following code 
```
Geolocator.getPositionStream(
          locationSettings: LocationSettings(
              accuracy: LocationAccuracy.bestForNavigation,
          )
      )
```

I have ignored the rest of the params for testing purpose. 

When I replaced the location settings with Apple Setting for iOS as suggested above, background location updates are working.

```
AppleSettings(
            accuracy: LocationAccuracy.bestForNavigation,
            allowBackgroundLocationUpdates: true,
            showBackgroundLocationIndicator: true,
            activityType: ActivityType.automotiveNavigation
        );
``` 

I have followed all the steps like allowing location updates in background mode, always location permission.

landyrev:
hey, @Faiizii, what's your `geolocator_apple` version? (you can find it in the `pubspec.lock`) 

Also, do you use the `getCurrentPosition` somewhere in your code? If yes, then it might be related to this issue: https://github.com/Baseflow/flutter-geolocator/pull/1169

Faiizii:
`geolocator_apple` is 2.2.3. No, I've just used the `getPositionStream` to get continuous updates.

mvanbeusekom:
Hi @landyrev, 

Reading the original comment from @Faiizii, it seems that the only problem is that the documentation doesn't explain to use the `AppleSettings` and specifically set the `allowBackgroundLocationUpdates` parameter to `true` if you'd want to enable background support for iOS.

He originally used the `LocationSettings` object which is mentioned in the documentation but doesn't turn on support for background services (as this is Apple specific).

landyrev:
@mvanbeusekom, ah, right. We set allowBackgroundLocationUpdates equal to `true` only when you use `AppleSettings`.

Yeah, fair enough, I'll draft a PR

Issue #581, opened 2020-11-05
Handling PermissionRequestInProgressException on iOS


kuhnroyal:
## 🐛 Bug Report
On iOS, when the app goes to background while the permission dialog is showing (due to pressing power button or display timeout, usually for 30+ seconds or so) and the phone is activated again, the permission dialog is gone. Afterwards it is not possible to request a permission again because it will always throw `PermissionRequestInProgressException`.

### Expected behavior
The plugin should detect this case and ensure that the permission can be requested again.
Somehow the `confirmationHandler` needs to be cleaned up here: https://github.com/Baseflow/flutter-geolocator/blob/master/geolocator/ios/Classes/Handlers/PermissionHandler.m#L37

### Reproduction steps
1. Make a call that requests the permission so that the dialog is showing (`Geolocator.getCurrentPosition()`)
2. Press the power button and let it sleep for 30+ seconds so that the app goes fully to background
3. Wake the phone again, the dialog should be gone
4. Repeat 1.

**Version:** 6.1.4

**Platform:** 
- [x] :iphone: iOS
- [ ] :robot: Android

Sempakonka:
Hi @kuhnroyal 

Thanks for filling this issue. I was able to reproduce it.

If you would like, I think Baseflow would love to see your PR!

<details>
<summary>flutter doctor -v</summary>

```
[✓] Flutter (Channel master, 1.25.0-5.0.pre.141, on macOS 11.0.1 20B50 darwin-x64, locale nl-NL)
    • Flutter version 1.25.0-5.0.pre.141 at /Users/sempakonka/Documents/flutter
    • Framework revision 36bed2b4c3 (13 hours ago), 2020-12-09 20:13:04 -0500
    • Engine revision 8518a5bbe6
    • Dart version 2.12.0 (build 2.12.0-135.0.dev)

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /Users/sempakonka/Library/Android/sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6915495)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 12.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
[...]

PR #1217, opened 2023-02-13
OpenIssue #581Handling PermissionRequestInProgressException on iOS fixed

Abisheik-IAN:
It return is removed from the ios native permissionhandle.m

In the permission request the popup appears and once the mobile locked then the popup disappear and the click will not do anything(thing is the await function waiting for the response from popup but the popup disappear(it is default iOS fn).

https://github.com/Baseflow/flutter-geolocator/issues/581

### DIFF
--- a/geolocator_apple/ios/Classes/Handlers/PermissionHandler.m
+++ b/geolocator_apple/ios/Classes/Handlers/PermissionHandler.m
@@ -54,7 +54,6 @@
     // Permission request is already running, return immediatly with error
     errorHandler(GeolocatorErrorPermissionRequestInProgress,
                  @"A request for location permissions is already running, please wait for it to complete before doing another request.");
-    return;
   }
   
   self.confirmationHandler = confirmationHandler;


geolocator_apple/ios/Classes/Handlers/PermissionHandler.m
```diff
@@ -54,7 +54,6 @@ - (void) requestPermission:(PermissionConfirmation)confirmationHandler
     // Permission request is already running, return immediatly with error
     errorHandler(GeolocatorErrorPermissionRequestInProgress,
                  @"A request for location permissions is already running, please wait for it to complete before doing another request.");
-    return;
```
Dolar1, line 57: Tried removing return and giving local ref locally but it still get's produced.

mvanbeusekom:
Closing this PR as this is not the correct way to resolve the issue. 

The error is caused when people dismis the permission dialog or put the application in the background while the permission dialog is showing. Unfortunately the iOS doesn't inform the geolocator (nor any other application) that the dialog is dismissed and therefore the geolocator keeps the `result` object active causing the error.

PR #1226, opened 2023-03-06
Add setOngoing flag to android notification options

flodaniel:
Adding a new flag `setOngoing` that sets the android notification ongoing ("persistent")

Reference: https://developer.android.com/reference/android/app/Notification.Builder.html#setOngoing(boolean)

Currently the notification can be dismissed by the user

With the flag enabled the notification cannot be dismissed by the user and must be dismissed by the system

No. The flag defaults to `false`

You can use these minimal `AndroidSettings`:
```dart
AndroidSettings(
        foregroundNotificationConfig: const ForegroundNotificationConfig(
          notificationText: 'notificationText',
          notificationTitle: 'notificationTitle',
          setOngoing: true,
        ),
      );
```


timbotimbo:
Some context from the [docs](https://developer.android.com/guide/components/foreground-services) why this is useful since API 33.
> ## User can dismiss notification by default
>
> Starting in Android 13 (API level 33), users can dismiss the notification associated with a foreground service by default. To do so, users perform a swipe gesture on the notification. Traditionally, the notification isn't dismissed unless the foreground service is either stopped or removed from the foreground.
>
> If you want the notification non-dismissable by the user, pass true into the [`setOngoing()`](https://developer.android.com/reference/android/app/Notification.Builder#setOngoing(boolean)) method when you create your notification using `Notification.Builder`.

mvanbeusekom (approved):
LGTM, thanks for this great contribution.

PR #1245, opened 2023-04-25
LocationManagerClient to avoid FUSED_PROVIDER

androidseb:
!!!WARNING!!! This fixed an issue for me but I don't have enough understanding of the overall code to be sure this isn't causing undesired side-effects - MERGE AT YOUR OWN RISK - I'm making this code change available in case it helps, along with a description of the bug it is fixing.

The current implementation results in the `getBestProvider` method sometimes returning `LocationManager.PASSIVE_PROVIDER`, which is not what I would expect from the LocationManagerClient, since we're trying to avoid using the fused provider.

This change is to ensure the provider returned is always `LocationManager.GPS_PROVIDER` or `LocationManager.NETWORK_PROVIDER`.

In my app using `forceLocationManager` everywhere, this creates the following bug:
* Start the app with location services disabled
* Attempt to access your location and get prompted to enable location services
* Enable location services
* Attempt to get the location from the app
* The bug happens:
    * EXPECTED: the location becomes available to the app
    * ACTUAL: the location is not available to the app until the app is restarted

I hope this helps!

### DIFF
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
@@ -75,43 +75,28 @@ class LocationManagerClient implements LocationClient, LocationListener {
 
   private static String getBestProvider(
       LocationManager locationManager, LocationAccuracy accuracy) {
-    Criteria criteria = new Criteria();
-
-    criteria.setBearingRequired(false);
-    criteria.setAltitudeRequired(false);
-    criteria.setSpeedRequired(false);
-
+    final boolean gpsEnabled = locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER);
+    final boolean networkEnabled = locationManager.isProviderEnabled(LocationManager.NETWORK_PROVIDER);
     switch (accuracy) {
       case lowest:
-        criteria.setAccuracy(Criteria.NO_REQUIREMENT);
-        criteria.setHorizontalAccuracy(Criteria.NO_REQUIREMENT);
-        criteria.setPowerRequirement(Criteria.NO_REQUIREMENT);
-        break;
       case low:
-        criteria.setAccuracy(Criteria.ACCURACY_COARSE);
-        criteria.setHorizontalAccuracy(Criteria.ACCURACY_LOW);
-        criteria.setPowerRequirement(Criteria.NO_REQUIREMENT);
-        break;
       case medium:
-        criteria.setAccuracy(Criteria.ACCURACY_COARSE);
-        criteria.setHorizontalAccuracy(Criteria.ACCURACY_MEDIUM);
-        criteria.setPowerRequirement(Criteria.POWER_MEDIUM);
-        break;
+        if (networkEnabled) {
+          return LocationManager.NETWORK_PROVIDER;
+        }else if (gpsEnabled){
+          return LocationManager.GPS_PROVIDER;
+        }else{
+          return LocationManager.PASSIVE_PROVIDER;
+        }
       default:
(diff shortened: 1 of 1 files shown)


androidseb:
I see... maybe I missed the point of the `LocationManagerClient` class, but my understanding was that this class is associated with the `forceLocationManager` parameter available in the dart API, so it should **never** use the fused location provider?

If `forceLocationManager` is not used, then the `FusedLocationClient` is used instead, that's what this other class is for, no?

Whoever has problems with `forceLocationManager` can stop using it, but at least they have the choice to use one or the other exclusively...

What I've found works pretty well for me is that I use forceLocationManager=true for Android < 10 and then forceLocationManager=false for Android 10 and up.

By the way, the code of this PR is a bit outdated, I ended up creating two listeners to listen both on the network provider and the GPS provider, would you like to see the code there? I don't mind sharing the code if you think it will be useful to you, but if you'd rather not change due to the risk, I understand. Unfortunately I don't have enough context or experience in this project to make educated recommendations, so I'll defer to you.

Please let me know if you'd like to look into those changes more, otherwise I don't mind closing this PR either.

Stypox:
As far as I understand, `FusedLocationClient` uses Google's fused client and strictly requires Google Play Services and a Google API key. But fortunately, many (all?) phones also allow accessing the "fused" client through the standard `LocationManagerClient`, which just uses Android APIs and does not require neither Google Play Services nor API keys.

YourFather25689 (changes requested):
Do it perfectly, If you don't know how to write a code then go to hell. Don't Waste people's time, From Now if you can change a file & create a pull request Then, I will expose you in front of ALL. UNDERSTAND NOW.

YourFather25689:
Do it perfectly, If you don't know how to write a code then go to hell. Don't Waste people's time, From Now if you can change a file & create a pull request Then, I will expose you in front of ALL. UNDERSTAND NOW.

YourFather25689:
Do it perfectly, If you don't know how to write a code then go to hell. Don't Waste people's time, From Now if you can change a file & create a pull request Then, I will expose you in front of ALL. UNDERSTAND NOW.

YourFather25689:
Do it perfectly, If you don't know how to write a code then go to hell. Don't Waste people's time, From Now if you can change a file & create a pull request Then, I will expose you in front of ALL. UNDERSTAND NOW.

mvanbeusekom:
Hi @androidseb,

Thank you for filing this PR and putting some time and trust into this plugin. Unfortunately I think you are a misguided by the naming of the different providers and classes. 

The `LocationManager.FUSED_PROVIDER` (introduces in Android API 31) only relies on the `LocationManager` and has no relation with Google Play Services. Specifying the `LocationManager.FUSED_PROVIDER` will instruct Android (or the `LocationManager` class) to combine all other available providers to get to the most optimal (battery consumption vs. accuracy) location result. 

Personally (I am unable to find information to back this up) I think the `LocationManager.FUSED_PROVIDER` is introduced to meet all those users that cannot use Google Play Services (e.g. in countries where it is disabled or forbidden) and still provide similar benefits.

The `FusedLocationProviderClient` comes with Google Play Services but relies on the `LocationManager` API. The implementation however (before `LocationManager.FUSED_PROVIDER` was introduced) had significant benefits on the power consumption of the device. It also added additional features, such as auto resolution when the location services are disabled (this is not supported by the `LocationManager` class.

I will close this PR for now, as I truly feel the current solution is the better option. However I am open for more feedback since the Android documentation is not very clear on the differences.

androidseb:
@mvanbeusekom thanks for your response.

The current solution might be the better option for most and I can't find any strong arguments to suggest otherwise so I agree closing this PR makes sense. There is that bug I described out there, but it's a bit of a niche case... Thanks for having a look!

PR #1264, opened 2023-06-01
Cancel location stream on `getCurrentPosition` timeout

### LINKED ISSUE #1259: Can't turn off location updates
## 🐛 Bug Report

**Version:** 4.1.7

**Platform:** 
- [ ] :robot: Android

I think the plugin needs a function to stop positioning, in android plugin, in some cases, we need to stop positioning manually. 

For example, startPositionUpdates has not returned a result for a long time, or the positioning has timed out. At this time, we cannot close the positioning, which causes the positioning to be permanent and consumes power.

JeroenWeener:
This PR fixes a bug where Android kept polling for a location after a request for the current location timed out.
Additionally, it enhances the `MethodChannelMock` class to cater to multiple mock methods.

Currently, when calling `getCurrentPosition()` with a provided `timeout`, the plugin will keep polling the gps on Android after the timeout, instead of cancelling the operation.

After a timeout, Android is instructed to stop polling the gps.

This PR does not introduce a breaking change.

The PR includes a test that checks whether the cancel request is fired after the request times out. The `MethodChannelMock` class was updated in the process.

Closes #1259.

### DIFF
--- a/geolocator_android/CHANGELOG.md
+++ b/geolocator_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.9
+
+* Fixes a bug where the location service would not stop correctly when the location stream timed out.
+
 ## 4.1.8
 
 * Adds compatibility with Android Gradle Plugin 8.0.
--- a/geolocator_android/CHANGELOG.md
+++ b/geolocator_android/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 4.1.9
 
-* Fixes a bug where the location service would not stop correctly when the location stream timed out.
+* Resolves an issue where the location service failed to stop properly upon timeout of the location stream.
 
 ## 4.1.8
 
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
@@ -7,6 +7,7 @@ import android.util.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 
 import com.baseflow.geolocator.errors.ErrorCodes;
 import com.baseflow.geolocator.errors.PermissionUndefinedException;
@@ -39,7 +40,7 @@ class MethodCallHandlerImpl implements MethodChannel.MethodCallHandler {
   private final GeolocationManager geolocationManager;
   private final LocationAccuracyManager locationAccuracyManager;
 
-  private final Map<String, LocationClient> pendingCurrentPositionLocationClients;
+  @VisibleForTesting final Map<String, LocationClient> pendingCurrentPositionLocationClients;
 
   @Nullable private Context context;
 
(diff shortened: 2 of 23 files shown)


geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
```diff
@@ -210,14 +225,17 @@ private void onGetCurrentPosition(MethodCall call, MethodChannel.Result result)
...
     Map<String, Object> map = (Map<String, Object>) call.arguments;
     boolean forceLocationManager = false;
-    if (map != null && map.get("forceLocationManager") != null) {
+    if (map.get("forceLocationManager") != null) {
       forceLocationManager = (boolean) map.get("forceLocationManager");
     }
     LocationOptions locationOptions = LocationOptions.parseArguments(map);
+    String requestId = (String) map.get("requestId");
+
     final boolean[] replySubmitted = {false};
```
JeroenWeener, line 234: This is likely an array so the boolean can be passed by reference, rather than by value. However, that does not seem necessary here.

geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
```diff
@@ -238,7 +257,24 @@ private void onGetCurrentPosition(MethodCall call, MethodChannel.Result result)
...
+      @SuppressWarnings("unchecked")
+      Map<String, Object> arguments = (Map<String, Object>) call.arguments;
+      String requestId = (String) arguments.get("requestId");
+
+      LocationClient locationClient = this.pendingCurrentPositionLocationClients.get(requestId);
+      if (locationClient != null) {
+        locationClient.stopPositionUpdates();
+      }
+
+      result.success(null);
```
JeroenWeener, line 278: Fail silently or call `result.error()`?

geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
```diff
@@ -238,7 +258,25 @@ private void onGetCurrentPosition(MethodCall call, MethodChannel.Result result)
...
+    Map<String, Object> arguments = (Map<String, Object>) call.arguments;
+    String requestId = (String) arguments.get("requestId");
+
+    LocationClient locationClient = this.pendingCurrentPositionLocationClients.get(requestId);
+    if (locationClient != null) {
+      locationClient.stopPositionUpdates();
+    }
+    this.pendingCurrentPositionLocationClients.remove(requestId);
+
+    result.success(null);
```
JeroenWeener, line 280: Fail silently in case there is no location client for the request identifier, or call result.error()?

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationMapper.java
```diff
@@ -6,7 +6,6 @@
 import java.util.HashMap;
 import java.util.Map;
 
-@SuppressWarnings("deprecation")
```
TimHoogstrate, line 9: Why is this un-deprecated? Or why was it deprecated? Any ideas?

JeroenWeener: VSCode indicated that this warning was redundant, so I removed it. I assumed that somewhere in a previous PR, a call to a deprecated method was removed/replaced, and the author forgot to remove this annotation.

geolocator_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.1.9
+
+* Fixes a bug where the location service would not stop correctly when the location stream timed out.
```
andreisas06, line 3:
```suggestion
* Resolves an issue where the location service failed to stop properly upon timeout of the location stream.
```

geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
mvanbeusekom: In this case I think failing silently would be the right approach. It should never happen and if for some reason it does happen returning an error will only cause confusion to the consuming developer.

geolocator_android/android/src/main/java/com/baseflow/geolocator/MethodCallHandlerImpl.java
JeroenWeener: Turns out we do need the array:

image

image

mvanbeusekom (approved):
LGTM!

PR #1271, opened 2023-06-07
Fix activity type propagation

RobertOdrowaz:
Bug fix

`activityType` in `AppleSettings` does nothing and location service is always started with `CLActivityType.other` type which is quite serious since it affects iOS power management. 

`activityType` is properly propagated to the native side.

No

### DIFF
--- a/geolocator_apple/CHANGELOG.md
+++ b/geolocator_apple/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.2.7
 
-* Fix activity type setting propagation.
+* Fixes the propagation of the activity type setting.
 
 ## 2.2.6
 
--- a/geolocator_apple/CHANGELOG.md
+++ b/geolocator_apple/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.2.7
+
+* Fix activity type setting propagation.
+
 ## 2.2.6
 
 * Ensures the `isLocationServicesEnabled` method is executed in a background thread.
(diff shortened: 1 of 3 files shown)


geolocator_apple/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.2.7
+
+* Fix activity type setting propagation.
```
mvanbeusekom, line 3: Entries in the changelog should use present tense indicative for verbs, with "this version" as an implied subject. Please update this entry to:
```suggestion
* Fixes the propagation of the activity type setting.
```

mvanbeusekom (approved):
Thank you very much for this bug fix. Changes look good modules a small style issue, would be nice if you could accept the suggestion.

RobertOdrowaz:
@mvanbeusekom Thanks for the review. 

What do you think about updating the dependency constraint in the main `geolocator` package and releasing a new version. I haven't included it in the PR because it seems you haven't done it in the past for platform package changes. 

I believe there are several advantages of updating the constraints:
1. Depending on a current version of the main package ensures that current version of platform packages are used
2. The main package changelog includes all of the fixes related to the package
3. Prevents the perception of the package being unmaintained

I can see only one reason not to do it: potential need to use older version of a specific platform package. However, this seems uncommon enough that a dependency override should be used if necessary.

mvanbeusekom:
Hi @RobertOdrowaz,

Since this is a patch fix it will automatically be picked up based on the dependency constraints in the `geolocator` package, so there is no need to release the `geolocator` package separately. 

The advantages you mention do hold true (option 1 in a less extend, as developers adding the `geolocator` package will always download the latest version of the `geolocator_apple` package based on the constraints specified). So if you are willing to create a separate PR to update the dependencies in the `geolocator` package it would be appreciated but not strictly necessary.

PR #1324, opened 2023-09-13
Fix inaccurate location readings

### LINKED ISSUE #1114: No position, when forceAndroidLocationManager: true and Google Play Services installed
## 🐛 Bug Report

It's not possible to retrieve a location, when `forceAndroidLocationManager: true` on a device with Google Play Services installed. Detailed findings below.

### Reproduction steps

I have written a minimal example `main.dart`. Just run it with different settings and see for yourself.

```dart
import 'package:flutter/material.dart';
import 'package:geolocator/geolocator.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: const MyHomePage(title: 'Geolocator Test'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({Key? key, required this.title}) : super(key: key);

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

/// Determine the current position of the device.
///
/// When the location services are not enabled or permissions
/// are denied the `Future` will return an error.
Future<Position> _determinePosition() async {
  bool serviceEnabled;
  LocationPermission permission;
  Position pos;

  // Test if location services are enabled.
  permission = await Geolocator.checkPermission();
  if (permission == LocationPermission.denied) {
    permission = await Geolocator.requestPermission();
    if (permission == LocationPermission.denied) {
      // Permissions are denied, next time you could try
      // requesting permissions again (this is also where
      // Android's shouldShowRequestPermissionRationale
      // returned true. According to Android guidelines
      // your App should show an explanatory UI now.
      return Future.error('Location permissions are denied');
    }
  }

  if (permission == LocationPermission.deniedForever) {
    // Permissions are denied forever, handle appropriately.
    return Future.error(
        'Location permissions are permanently denied, we cannot request permissions.');
  }

  // When we reach here, permissions are granted and we can
  // continue accessing the position of the device.
  print('Permission granted, asking for current location.');
  pos = await Geolocator.getCurrentPosition(
    forceAndroidLocationManager: true
  );
  print(pos);
  return pos;
}

class _MyHomePageState extends State<MyHomePage> {
  Position? _pos;

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        // Here we take the value from the MyHomePage object that was created by
        // the App.build method, and use it to set our appbar title.
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            const Text(
              'You are here:',
            ),
            Text(
              '$_pos',
              style: Theme.of(context).textTheme.headline4,
            ),
          ],
        ),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () async {
          _pos = await _determinePosition();
          setState(() {});
          },
        tooltip: 'Increment',
        child: const Icon(Icons.add),
      ), // This trailing comma makes auto-formatting nicer for build methods.
    );
  }
}
```
In `build.gradle` 
```
android {
    compileSdkVersion 33
    ...
    }
```
    
In `AndroidManifest.xml`
```
    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
```

My findings are as follows:

Device **with** Google Play Services

- When **forceAndroidLocationManager=true** and **LocationService=off** and **GoogleLocationAccuracy=off** => _Unhandled Exception: The location service on the device is disabled._

- When **forceAndroidLocationManager=true** and **LocationService=off** and **GoogleLocationAccuracy=on** => _Unhandled Exception: The location service on the device is disabled._

- When **forceAndroidLocationManager=true** and **LocationService=on** and **GoogleLocationAccuracy=off** => _No Position, No Error Message) [Here lays the PROBLEM!!!]_

- When **forceAndroidLocationManager=true** and **LocationService=on** and **GoogleLocationAccuracy=on** => _No Position, No Error Message) [Here lays the PROBLEM!!!]_

Device **without** Google Play Services

- When **forceAndroidLocationManager=true** and **LocationService=off** => _Unhandled Exception: The location service on the device is disabled._

- When **forceAndroidLocationManager=true** and **LocationService=on** => _POSITION_

I hope someone can confirm my findings or even can provide a (quick) fix.

### Configuration

**Version:** 8.2.1 / 9.0.1.

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

JeroenWeener:
Users have reported that locations have an wildly low accuracy on Android 13 since `geolocator_android: 4.2.3`. This seems to be caused by the fact that a different location provider is returned from `getBestProvider()`. Instead of the fused or gps provider, the network provider is used. To remedy this, this PR replaces the use of `getBestProvider()` with a custom implementation that returns the most accurate location provider. Note that `getBestProvider()` will be deprecated starting from API 34 anyway, with no clear replacement. Additionally, the `LocationManagerCompat` class and its associates are used to indicate a `LocationRequestCompat.Quality` to the OS.

Closes #1114.

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
@@ -153,20 +153,24 @@ class LocationManagerClient implements LocationClient, LocationListenerCompat {
     this.positionChangedCallback = positionChangedCallback;
     this.errorCallback = errorCallback;
 
-    this.currentLocationProvider = getBestProvider(this.locationManager);
-
-    if (this.currentLocationProvider == null) {
-      errorCallback.onError(ErrorCodes.locationServicesDisabled);
-      return;
-    }
-
+    LocationAccuracy accuracy = LocationAccuracy.best;
     long timeInterval = 0;
     float distanceFilter = 0;
     @LocationRequestCompat.Quality int quality = LocationRequestCompat.QUALITY_BALANCED_POWER_ACCURACY;
     if (this.locationOptions != null) {
       timeInterval = locationOptions.getTimeInterval();
       distanceFilter = locationOptions.getDistanceFilter();
-      quality = accuracyToQuality(locationOptions.getAccuracy());
+      accuracy = locationOptions.getAccuracy();
+      quality = accuracyToQuality(accuracy);
+    }
+
+    this.currentLocationProvider = accuracy == LocationAccuracy.lowest
+        ? LocationManager.PASSIVE_PROVIDER
+        : getBestProvider(this.locationManager);
+
+    if (this.currentLocationProvider == null) {
+      errorCallback.onError(ErrorCodes.locationServicesDisabled);
+      return;
     }
 
     final LocationRequestCompat locationRequest = new LocationRequestCompat.Builder(timeInterval)
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
@@ -4,22 +4,24 @@ import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
-import android.location.Criteria;
 import android.location.Location;
-import android.location.LocationListener;
 import android.location.LocationManager;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Looper;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.core.location.LocationListenerCompat;
+import androidx.core.location.LocationManagerCompat;
+import androidx.core.location.LocationRequestCompat;
 
 import com.baseflow.geolocator.errors.ErrorCallback;
 import com.baseflow.geolocator.errors.ErrorCodes;
 
 import java.util.List;
 
-class LocationManagerClient implements LocationClient, LocationListener {
+class LocationManagerClient implements LocationClient, LocationListenerCompat {
 
   private static final long TWO_MINUTES = 120000;
   private final LocationManager locationManager;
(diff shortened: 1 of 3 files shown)


geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
```diff
@@ -73,45 +75,35 @@ static boolean isBetterLocation(Location location, Location bestLocation) {
...
+          return enabledProviders.get(0);
+      } else {
+          return null;
+      }
+  }
 
-    return provider;
+  private static @LocationRequestCompat.Quality int accuracyToQuality(@NonNull LocationAccuracy accuracy) {
+      switch (accuracy) {
+          case lowest:
```
mvanbeusekom, line 96: I think `lowest` should return `LocationRequestCompat.PASSIVE_INTERVAL`:

```suggestion
          case lowest:
                  return LocationRequestCompat.PASSIVE_INTERVAL;
```

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
```diff
@@ -73,45 +75,35 @@ static boolean isBetterLocation(Location location, Location bestLocation) {
...
-        break;
-    }
-
-    String provider = locationManager.getBestProvider(criteria, true);
-
-    if (provider.trim().isEmpty()) {
-      List<String> providers = locationManager.getProviders(true);
-      if (providers.size() > 0) provider = providers.get(0);
-    }
+  private static @Nullable String getBestProvider(@NonNull LocationManager locationManager) {
```
mvanbeusekom, line 78: I think we should pass in the `LocationAccuracy` as well and if the supplied value equals `LocationAccuracy.lowest` immediately return `LocationManager.PASSIVE_PROVIDER`:

```suggestion
  private static @Nullable String getBestProvider(@NonNull LocationManager locationManager, @NonNull LocationAccuracy accuracy) {
      if (accuracy == LocationAccuracy.lowest) {
          return LocationManager.PASSIVE_PROVIDER;
      }
      
```

mvanbeusekom (changes requested):
Looks good however, I think we should also handle the `PASSIVE_PROVIDER` when the support `LocationAccuracy` is set to `lowest`. See also my comments.

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
JeroenWeener: I agree with using the `PASSIVE_PROVIDER` on `lowest`. I did not put the logic in this method, as I feel it does not belong there. Instead, I added it to `startPositionUpdates`.

mvanbeusekom: I think it should be in this method, there should only be one place which determines the provider. This will ensure we always use the same logic when we need to determine the provider on multiple places. 

Maybe we should however rename the method as I agree that the name `getBestProvider` doesn't fit well with the `PASSIVE_PROVIDER`. So lets rename it to `getProvider` or `determineProvider`.

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
mvanbeusekom: Why do you think this is not a quality parameter? According to the [documentation](https://developer.android.com/reference/androidx/core/location/LocationRequestCompat#PASSIVE_INTERVAL()) the `LocationRequestCompat.PASSIVE_PROVIDER` is a valid quality parameter with the const value  of `9223372036854775807 `.

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
```diff
@@ -161,27 +153,40 @@ public void startPositionUpdates(
...
     if (this.locationOptions != null) {
       timeInterval = locationOptions.getTimeInterval();
       distanceFilter = locationOptions.getDistanceFilter();
+      accuracy = locationOptions.getAccuracy();
+      quality = accuracyToQuality(accuracy);
+    }
+
+    this.currentLocationProvider = accuracy == LocationAccuracy.lowest
+        ? LocationManager.PASSIVE_PROVIDER
+        : getBestProvider(this.locationManager);
```
mvanbeusekom, line 169: The logic to determine the provider should al be encapsulated in one method (as mentioned in my other comment). Splitting it up like this will increase the change of mistakes when the provider needs to be determined on multiple places in the code.

mvanbeusekom (changes requested):

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
mvanbeusekom: You are correct, I miss read the documentation. This should be good the way it is.

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/LocationManagerClient.java
```diff
@@ -161,27 +153,40 @@ public void startPositionUpdates(
...
-
+    LocationAccuracy accuracy = LocationAccuracy.best;
     long timeInterval = 0;
     float distanceFilter = 0;
+    @LocationRequestCompat.Quality int quality = LocationRequestCompat.QUALITY_BALANCED_POWER_ACCURACY;
     if (this.locationOptions != null) {
       timeInterval = locationOptions.getTimeInterval();
       distanceFilter = locationOptions.getDistanceFilter();
+      accuracy = locationOptions.getAccuracy();
+      quality = accuracyToQuality(accuracy);
```
mvanbeusekom, line 164: Rereading the [documentation](https://developer.android.com/reference/androidx/core/location/LocationRequestCompat#PASSIVE_INTERVAL()), we should use the `PASSIVE_INTERVAL` here when the `LocationAccuracy` is set to lowest:

```suggestion
      accuracy = locationOptions.getAccuracy();
      distanceFilter = locationOptions.getDistanceFilter();
      quality = accuracyToQuality(accuracy);
      timeInterval = (accuracy == LocationAccurracy.lowest) 
          ? LocationRequestCompat.PASSIVE_INTERVAL 
          : locationOptions.getTimeInterval();
```

PR #1343, opened 2023-09-26
Suppress deprecation warning on Android build

JeroenWeener:
The `LocationManager` overrides the method `onStatusChanged`, which is deprecated. This triggers a warning during the Android build process when using the geolocator plugin. According to the documentation, this method is only triggered on Android Q and below, and it is deprecated since Android Q. Therefore, we need to keep overriding this method to support older Android version, but can safely suppress the warning for later versions.

The linter marks the deprecation suppression an unneeded. The build process does trigger the deprecation warning, though. Therefore, a second suppression was added, to suppress the warning that the deprecation suppression is unneeded.

More context is available in the Android documentation at https://developer.android.com/reference/android/location/LocationListener#onStatusChanged(java.lang.String,%20int,%20android.os.Bundle).

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/geolocator_android/CHANGELOG.md
+++ b/geolocator_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.3.1
+
+- Suppresses a deprecation warning for `LocationListenerCompat.onStatusChanged` when building for Android.
+
 ## 4.3.0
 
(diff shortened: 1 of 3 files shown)


mvanbeusekom (approved):
LGTM

PR #1351, opened 2023-10-03
Fixes- Issues when native location dialog closes after pressing the power button

Dolar1:
*Removed exception as not needed we are already setting confirmationHandler to nil in the end*

1. https://github.com/Baseflow/flutter-geolocator/issues/581

## Pre-launch Checklist

- [✅] I made sure the project builds.
- [✅] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [✅] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is does not need version changes.
- [✅] I updated `CHANGELOG.md` to add a description of the change.
- [✅] I updated/added relevant documentation (doc comments with `///`).
- [✅] I rebased onto `main`.
- [✅] I added new tests to check the change I am making, or this PR does not need tests.
- [✅] I made sure all existing and new tests are passing.
- [✅] I ran `dart format .` and committed any changes.
- [✅] I ran `flutter analyze` and fixed any errors.

### DIFF
--- a/geolocator_apple/ios/Classes/Handlers/PermissionHandler.m
+++ b/geolocator_apple/ios/Classes/Handlers/PermissionHandler.m
@@ -50,13 +50,6 @@
     return;
   }
   
-  if (self.confirmationHandler) {
-    // Permission request is already running, return immediatly with error
-    errorHandler(GeolocatorErrorPermissionRequestInProgress,
-                 @"A request for location permissions is already running, please wait for it to complete before doing another request.");
-    return;
-  }
-  
   self.confirmationHandler = confirmationHandler;
   self.errorHandler = errorHandler;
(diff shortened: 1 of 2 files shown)


TimHoogstrate:
Dear @Dolar1,

Thank you for submitting your PR and helping us improving the packages. I've noticed that this PR is open for a while now. I've discussed your PR this within the team and it looks like it is only removing the exception. However, things get dirty when we omit the exception and let the process continue. A new `FlutterResult` will be generated and overrides the old `FlutterResult`. A result on the old request will then never be handled properly. And thus it will introduce another issue. However, we are planning to update this in the future when we introduce the a new instance manager. Therefore, I'll close this PR for now. Feel free to add information so that we can use it in the future.

Kind regards

PR #1361, opened 2023-10-30
Implements `color` parameter in `ForegroundNotifcationOptions`

### LINKED ISSUE #1277: Custom Android notification icon color
## 🚀 Feature Requests

Right now, only the notification title, message, and icon can be customized. It would be great if we could also customize the icon's (background) color, instead of being forced to use the system's default.

### Contextualize the feature
The framework creates the Android `NotificationBuilder`, but doesn't expose many of the available methods to Flutter.

### Platforms affected (mark all that apply)
- [ ] :iphone: iOS
- [x] :robot: Android

JeroenWeener:
Implements `color` parameter in `ForegroundNotificationOptions`.

Closes #1277.

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/BackgroundNotification.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/location/BackgroundNotification.java
@@ -24,10 +24,10 @@ public class BackgroundNotification {
     private NotificationCompat.Builder builder;
 
     public BackgroundNotification(
-            Context context,
-            String channelId ,
-            Integer notificationId,
-            ForegroundNotificationOptions options
+        @NonNull Context context,
+        @NonNull String channelId ,
+        @NonNull Integer notificationId,
+        ForegroundNotificationOptions options
     ) {
         this.context = context;
         this.notificationId = notificationId;
@@ -87,6 +87,12 @@ public class BackgroundNotification {
                 .setContentText(options.getNotificationText())
                 .setContentIntent(buildBringToFrontIntent())
                 .setOngoing(options.isSetOngoing());
+
+        int notificationColor = options.getColor();
+        if (notificationColor != -1) {
+            builder = builder
+                .setColor(options.getColor());
+        }
 
         if (notify) {
             NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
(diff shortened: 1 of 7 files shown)


geolocator_android/android/src/main/java/com/baseflow/geolocator/location/BackgroundNotification.java
```diff
@@ -88,6 +88,12 @@ private void updateNotification(
                 .setContentIntent(buildBringToFrontIntent())
                 .setOngoing(options.isSetOngoing());
 
+        int notificationColor = options.getColor();
+        if (notificationColor != -1) {
```
TimHoogstrate, line 92: Where has this magical -1 evolved from? Maybe it is clearer if you add a "DEFAULT" or "NOT_SET" constant or something both on Dart as on Java Side or add an inline documentation of the -1.

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/BackgroundNotification.java
```diff
@@ -88,6 +88,12 @@ private void updateNotification(
...
+            builder = builder
+                .setColor(options.getColor());
```
mvanbeusekom, line 94: Seems more logical to use the `notificationColor` here:

```suggestion
                .setColor(notificationColor);
```

mvanbeusekom (changes requested):
Looks good, one small remark

geolocator_android/lib/src/types/foreground_settings.dart
```diff
@@ -51,16 +58,34 @@ class ForegroundNotificationConfig {
   /// the user cannot dismiss it
   final bool setOngoing;
 
+  /// Accent color (an ARGB integer like the constants in Color) to be applied
+  /// by the standard Style templates when presenting this notification.
+  ///
+  /// The current template design constructs a colorful header image by
+  /// overlaying the icon image (stenciled in white) atop a field of this color.
+  /// Alpha components are ignored.
+  final Color? color;
```
TimHoogstrate, line 67: What happens if this remains null?

TimHoogstrate (approved):
Just some little remarks to clarify to the users/future developers.

geolocator_android/android/src/main/java/com/baseflow/geolocator/location/BackgroundNotification.java
JeroenWeener: Fair point. I made the integer nullable so we can check if the parameter was supplied that way.

mvanbeusekom (approved):
LGTM

PR #1367, opened 2023-11-06
Enhancement/1366

TimHoogstrate:
*Replace this paragraph with a short description of what issue this pull request (PR) solves and provide a description of the change.  Consider including before/after screenshots.*

## Pre-launch Checklist

- [ ] I made sure the project builds.
- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I rebased onto `main`.

### DIFF
--- a/geolocator_android/CHANGELOG.md
+++ b/geolocator_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.4.1
+
+- Adds addOnFailureListener to getSettingsClient to prevent crashes.
+
 ## 4.4.0
 
 - Adds `color` to `ForegroundNotificationConfig` to set the color of the notification icon.
--- a/geolocator_android/android/src/main/java/com/baseflow/geolocator/errors/ErrorCodes.java
+++ b/geolocator_android/android/src/main/java/com/baseflow/geolocator/errors/ErrorCodes.java
@@ -4,6 +4,7 @@ public enum ErrorCodes {
   activityMissing,
   errorWhileAcquiringPosition,
   locationServicesDisabled,
+    locationServicesFailed,
   permissionDefinitionsNotFound,
   permissionDenied,
   permissionRequestInProgress;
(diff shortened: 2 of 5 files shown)


greg-tekai-com:
Wow, great turnaround time!

Quick question as I'm not sure.   By adding the `addOnFailureListener`, should the `addOnCompleteListener` above it be changed to `addOnSuccessListener` so that it doesn't still catch failures, or will only the failure listener fire?  

I'm not that familiar with the code, but based on the docs, I was expecting `addOnCompleteListener` was like a `finally` block.  If not, please ignore me. 

Thanks!

jozefvodicka:
Thanks so much for proposed solution, is it possible to finalise it so the pull request can be created?

tentenponce:
hi @TimHoogstrate the crash points to `LocationSettingsResponse lsr = response.getResult();` on `FusedLocationClient` line 162. The failure handling is applied on `LocationServices.getSettingsClient(context).checkLocationSettings(new LocationSettingsRequest.Builder().build())` which seems to be successful since we are reaching the `response.getResult()` and encountering the crash from there.

I don't think the solution will fix the issue.

attaching also another similar issue with react native, and their approach resolves the issue. However, still it doesn't fix the root cause since they just added a simple try catch. This probably avoid crashes, while waiting for Google to fix it.
https://github.com/Agontuk/react-native-geolocation-service/pull/147/files

greg-tekai-com:
The problem is it still runs the code after the `if (!response.isSuccessful())` block on failures, leading to checking the result regardless.  It should exit the method after calling `listener.onLocationServiceError` in that block if it was a failure, or wrap the rest in an else block.

If we add the `addOnFailureListener` as per the PR, then it also needs to change the `addOnCompleteListener`  to `addOnSuccessListener`, removing the check as it should be redundant, or leave it, but ensure we don't call `listener.onLocationServiceError` again if it was a failure.  Calling it twice returns a result on the method channel twice and that will also fail.

Not sure we'll use it, but the new error code is nice addition so that we can determine it failed to check, rather than default to disabled.

kazbek-sultanov-t:
We are encountering exactly the same error. 
Can we continue on this and merge the proposed PR by @tentenponce ?

PR #1458, opened 2024-03-14
[Windows] Handle errors in request access

daniel-kane-everbridge-com:
Fixes crash under Windows when RequestAccess is called while the Geolocation Service is disabled. 
 - this is done by swallowing any error thrown from geolocator.RequestAccessAsync();


mvanbeusekom (approved):
LGTM, thank you for this contribution.

PR #1475, opened 2024-04-02
migrates geolocator_web to package:web

josh-burton:
This PR migrates from dart:html to package:web and js_interop.

A few notes/questions:
- I've increased the dart/flutter sdk versions in geolocator_web in order to use js_interop. I'm unsure if we need to bump these versions in the other geolocator packages as well.
- Mockito can no longer be used with the new js_interop types (at least I ran into errors with it being unable to generate mocks).
- The package:web geolocation methods now require a timeout value, rather than allowing null. Infinity is the value for timeouts in the html geolocation api, but passing double.infinity.toInt as timeout throws an error that this is an invalid value when invoked in javascript. Instead of defaulted to a timeout of 1 day, but this needs more input.
- the test now require running with `flutter test --platform chrome`, otherwise errors are thrown as js_interop can't be used in the dart vm.

## Pre-launch Checklist

- [x] I rebased onto `main`.


geolocator_web/lib/src/html_permissions_manager.dart
```diff
@@ -23,12 +22,10 @@ class HtmlPermissionsManager implements PermissionsManager {
...
-      _permissionQuery,
-    );
+    // navigator.permissions.query({ name: "geolocation" })
+    final web.PermissionStatus? status =
+        await _permissions?.query(_permissionQuery.jsify() as JSObject).toDart;
 
-    return status.state != null
-        ? toLocationPermission(status.state)
-        : LocationPermission.denied;
+    return toLocationPermission(status?.state);
```
mvanbeusekom, line 29: This will result in an exception when the `status` parameter is `null` (this might happen when the user dismisses the permission dialog without specifically granting or denying the permission). 

In these cases the permission request should always return the `denied` permission, informing the user to request permission again.

I think it would be good to use a similar construct as the old statement:
```suggestion
    return status != null
            ? toLocationPermission(status.state)
            : LocationPermission.denied;
```

geolocator_web/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 4.0.0
+
+**BREAKING CHANGE:**
+- Migrates from dart:html to package:web and dart:js_interop
+- Migrates to Dart SDK 3.3.0 and Flutter 3.16.0.
```
mvanbeusekom, line 5: nit: officially the bullets should be indented with two spaces for the correct markdown behaviour:
```suggestion
  - Migrates from dart:html to package:web and dart:js_interop
  - Migrates to Dart SDK 3.3.0 and Flutter 3.16.0.
```

geolocator_web/lib/src/html_geolocation_manager.dart
```diff
@@ -1,46 +1,77 @@
...
+          timeout:
+              timeout?.inMicroseconds ?? const Duration(days: 1).inMilliseconds,
+        ),
       );
-
-      return toPosition(geoPosition);
-    } on html.PositionError catch (e) {
-      throw convertPositionError(e);
+    } catch (e) {
+      completer.completeError(e);
```
mvanbeusekom, line 39: Please convert the error to one of the custom exceptions defined in the `convertPositionError` method before completing the completer. 

This will throw a known exception to the application developers so they don't have to update their application logic and handle unexpected exceptions solely for the geolocator_web plugin.

mvanbeusekom (changes requested):
Hi @josh-burton,

Sorry for the delay and thank you for creating this PR. In general the changes look good, I have made a few small remarks which would be great if you can address those.

geolocator_web/lib/src/html_geolocation_manager.dart
josh-burton: I've returned a platform exception so the caught exception can be passed to the detail field. Another option could be the `PositionUpdateException` exception

josh-burton:
> Hi @josh-burton,
> 
> Sorry for the delay and thank you for creating this PR. In general the changes look good, I have made a few small remarks which would be great if you can address those.

Thanks for the review. I believe I've addressed them now.

mvanbeusekom:
When merged solves #1513

geolocator_web/lib/src/html_geolocation_manager.dart
josh-burton: ok, my only concern there is the underlying exception will be lost. Perhaps we could add an optional cause field to the PositionUpdateException?

josh-burton:
@mvanbeusekom I see the workflows are failing. I had bumped the geolocator_web version to 4.0 as its a breaking change, but as 4.0 is not published the workflows now error.

What's the correct fix here?

mvanbeusekom:
The best would be to split the changes into 2 PRs. One updating the geolocator_web package and a second bumping the version for the geolocator package.

- However in this case the change in the geolocator package is so small that I don't mind the failing build. As long as the steps for the geolocator_web are green.

mvanbeusekom (approved):
LGTM

Issue #1171, opened 2022-11-09
forceLocationManager = true may still default to fused location


jbxbergdev:
When setting `forceLocationManager: true` in `AndroidSettings`, geolocator may still request fused location provider.

The reason: `LocationManagerClient` calls [`LocationManager.getBestProvider()`](https://developer.android.com/reference/android/location/LocationManager#getBestProvider(android.location.Criteria,%20boolean)) in order to get an appropriate location provider. This method may return [`LocationManager.FUSED_PROVIDER`](https://developer.android.com/reference/android/location/LocationManager#FUSED_PROVIDER) on certain devices (in my case Samsung A13, Android 12).

There should be a way to force usage of [`LocationManager.GPS_PROVIDER`](https://developer.android.com/reference/android/location/LocationManager#GPS_PROVIDER) if network location is not desired. Suggestion: Don't call `LocationManager.getBestProvider()`, but always use `LocationManager.GPS_PROVIDER`, when `forceLocationManager` is set to `true`.

mvanbeusekom:
Hi @jbxbergdev ,

Thank you for filing this issue and putting some time and trust into this plugin. Unfortunately I think you are a misguided by the naming of the different providers and classes. 

The `LocationManager.FUSED_PROVIDER` (introduces in Android API 31) only relies on the `LocationManager` and has no relation with Google Play Services. Specifying the `LocationManager.FUSED_PROVIDER` will instruct Android (or the `LocationManager` class) to combine all other available providers to get to the most optimal (battery consumption vs. accuracy) location result. 

Personally (I am unable to find information to back this up) I think the `LocationManager.FUSED_PROVIDER` is introduced to meet all those users that cannot use Google Play Services (e.g. in countries where it is disabled or forbidden) and still provide similar benefits.

The `FusedLocationProviderClient` comes with Google Play Services but relies on the `LocationManager` API. The implementation however (before `LocationManager.FUSED_PROVIDER` was introduced) had significant benefits on the power consumption of the device. It also added additional features, such as auto resolution when the location services are disabled (this is not supported by the `LocationManager` class.

I will close this issue for now, as I truly feel the current solution is correct. However I am open for more feedback since the Android documentation is not very clear on the differences.

jbxbergdev:
@mvanbeusekom the motivation behind my request was actually not the relation to Google Play services, but that my app needs to use location strictly from GNSS. Fused location also takes wifi and cellular networks into consideration. 
[...]

PR #1613, opened 2024-11-28
[geolocator_android] Introduce option to force a specific location provider   Issue/1171

jbxbergdev:
Added an option `AndroidSettings.forceProvider` that forces the plugin to use a specific provider out of the currently four available Android location providers. As discussed with @mvanbeusekom, this option is to be used in conjunction with `forceLocationManager = true`.

Example use case: An app needs to ensure locations are derived exclusively from GNSS and hence has to make sure no other sources like mobile networks are used.

#1171 

## Pre-launch Checklist

- [ ] I made sure the project builds.
- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I rebased onto `main`.

### DIFF
--- a/geolocator/lib/geolocator.dart
+++ b/geolocator/lib/geolocator.dart
@@ -8,6 +8,7 @@ import 'package:geolocator_platform_interface/geolocator_platform_interface.dart
 export 'package:geolocator_android/geolocator_android.dart'
     show
         AndroidSettings,
+        AndroidLocationProvider,
         ForegroundNotificationConfig,
         AndroidResource,
         AndroidPosition;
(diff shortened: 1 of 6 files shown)


jbxbergdev:
@mvanbeusekom not really sure how to get rid of the analyzer errors the proper way. Please advise, thanks.

TimHoogstrate:
@jbxbergdev,

Thanks for the contribution. I cannot see the build log anymore... Is it possible to rebase onto main and then we can have a look at the build logs. Maybe the analyser warnings are resolved recently.

Kind regards,

TimHoogstrate (commented):
Nice work! I added some comments. Can we also add the feature to the android package example app so that we can easily test the feature?

PR #1740, opened 2025-07-24
Fix: Add POST_NOTIFICATIONS permission

Rakibul25:
This PR adds the 'android.permission.POST_NOTIFICATIONS' permission to the Android manifest. This is required for apps targeting API 33 (Android 13) and above to use NotificationManagerCompat.notify without build failures related to missing permissions.

### DIFF
--- a/geolocator_android/android/src/main/AndroidManifest.xml
+++ b/geolocator_android/android/src/main/AndroidManifest.xml
@@ -1,11 +1,13 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-  package="com.baseflow.geolocator">
+    package="com.baseflow.geolocator">
+
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
 
     <application>
         <service
-                android:enabled="true"
-                android:exported="false"
-                android:foregroundServiceType="location"
-                android:name=".GeolocatorLocationService"/>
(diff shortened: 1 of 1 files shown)


mvanbeusekom:
Hi @Rakibul25,

We prefer not to add permissions to the `AndroidManifest.xml` file of the plugin it self. As these permissions will be merged into the `AndroidManifest.xml` file of the consuming app. Meaning that apps the use the permission_handler but don't require the `POST_NOTIFICATION` permission will end up having it declared in their final `AndroidManifest.xml` file after compilation. 

This should be an explicit choice of the app developer and therefore the app developer should manually add the permission to the `AndroidManifest.xml` file associated with their application.

Rakibul25:
Thank you for the clear explanation regarding plugin-level permissions. I completely understand and agree that it makes sense to keep permission declarations explicitly at the app level.
For clarity to other users, it's worth highlighting that developers will encounter build failures or runtime issues related to POST_NOTIFICATIONS if foreground services trigger notifications. They'll need to manually add <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/> to their app's AndroidManifest.xml and handle runtime permission requests. I've seen many users encountering "panic-inducing" lint errors during Gradle builds due to this.
Given this, I agree the solution lies with the app developer. You can close this issue. Thanks again!

Issue #1518, opened 2024-05-23
[Bug]: requestPermission Future never resolves, when application goes to background, when "Allow to use location" popup is displayed


tobolski:
### Please check the following before submitting a new issue.

- [X] I have searched the [existing issues](https://github.com/baseflow/flutter-geolocator/issues).
- [X] I have carefully [read the documentation](https://github.com/Baseflow/flutter-geolocator/blob/main/geolocator/README.md) and verified I have added the required platform specific configuration.

### Please select affected platform(s)

- [ ] Android
- [X] iOS
- [ ] Linux
- [ ] macOS
- [ ] Web
- [ ] Windows

### Steps to reproduce

1. Enable location on device and in application.
2. Execute: `final permission = await Geolocator.requestPermission();`
3. "Allow to use location" popup should appear:
![image](https://github.com/Baseflow/flutter-geolocator/assets/13032980/7f8f6bb5-c502-4f05-a65e-3868af92439a)
4. Lock the phone.
5. Unlock the phone.

### Expected results

- Bring back "Allow to use location" when going back to application.
OR
- Complete the Future with proper state.

### Actual results

1. "Allow to use location" is gone.
2. `final permission = await Geolocator.requestPermission();` is waiting forever.
3. Another call to this method results in exception:
```
Unhandled Exception: A request for location permissions is already running, please wait for it to complete before doing another request.
```

### Code sample

<details><summary>Code sample</summary>

```dart
final permission = await Geolocator.requestPermission();
```

</details>

### Screenshots or video

final permission = await Geolocator.requestPermission();

### Version

12.0.0

### Flutter Doctor output

<details><summary>Doctor output</summary>

```console
[✓] Flutter (Channel stable, 3.19.6, on macOS 14.4.1 23E224 darwin-arm64, locale en-PL)
    • Flutter version 3.19.6 on channel stable at /Users/xxx/git/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 54e66469a9 (5 weeks ago), 2024-04-17 13:08:03 -0700
    • Engine revision c4cd48e186
    • Dart version 3.3.4
    • DevTools version 2.31.1
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/xxx/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
[...]

PR #1774, opened 2026-03-12
fix(iOS): resolve permission dialog hang when app goes to background.

guide-linhnv:
*Replace this paragraph with a short description of what issue this pull request (PR) solves and provide a description of the change.  Consider including before/after screenshots.*

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/geolocator_apple/darwin/geolocator_apple/Sources/geolocator_apple/Handlers/PermissionHandler.m
+++ b/geolocator_apple/darwin/geolocator_apple/Sources/geolocator_apple/Handlers/PermissionHandler.m
@@ -8,12 +8,16 @@
 #import "../include/geolocator_apple/Handlers/PermissionHandler.h"
 #import "../include/geolocator_apple/Constants/ErrorCodes.h"
 #import "../include/geolocator_apple/Utils/PermissionUtils.h"
+#if TARGET_OS_IOS
+#import <UIKit/UIKit.h>
+#endif
 
 @interface PermissionHandler() <CLLocationManagerDelegate>
 
 @property (strong, nonatomic) CLLocationManager *locationManager;
 @property (strong, nonatomic) PermissionConfirmation confirmationHandler;
 @property (strong, nonatomic) PermissionError errorHandler;
+@property (assign, nonatomic) BOOL isWaitingForPermission;
 
 @end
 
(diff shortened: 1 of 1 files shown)


guide-linhnv:
## Summary                                                                                                                                                                  
                  
  This PR fixes issue #1518 where `Geolocator.requestPermission()` Future hangs indefinitely when the iOS app goes to background while the location permission dialog is
  displayed.

  ## Problem

  On iOS, when the location permission dialog is showing and the user:
  - Locks their phone
  - Switches to another app
  - Triggers any action that backgrounds the app

  iOS automatically dismisses the permission dialog **without** calling the `CLLocationManagerDelegate` callback. This causes:
  1. The `requestPermission()` Future to never complete (hang forever)
  2. Subsequent permission requests to fail with: *"A request for location permissions is already running"*

  ## Root Cause

  iOS's `CLLocationManager` delegate method `locationManager:didChangeAuthorizationStatus:` is not called when the permission dialog is dismissed due to app backgrounding.
  The Flutter side keeps waiting for a response that will never arrive.

  ## Solution

  Added app lifecycle monitoring to detect when the app returns to foreground after being backgrounded during a permission request:

  1. **Register for `UIApplicationDidBecomeActiveNotification`** when starting a permission request
  2. **Track permission request state** with `isWaitingForPermission` flag
  3. **On app becoming active**:
     - Wait 0.3s to allow any pending system callbacks to process
     - Check if still waiting for permission response
     - Resolve the callback with current authorization status
  4. **Proper cleanup** of notification observers in both `cleanUp` and `dealloc`

  ## Changes

  - `PermissionHandler.m`: Added lifecycle observer and handling logic (iOS only, wrapped in `#if TARGET_OS_IOS`)

  ## Testing

  - [x] Tested on iOS 17+ simulator and device
  - [x] Verified permission dialog works normally when user makes a choice
  - [x] Verified Future resolves correctly when app is backgrounded during dialog
  - [x] Verified subsequent permission requests work after backgrounding
  - [x] Verified no impact on macOS (changes are iOS-specific)

  ## Breaking Changes

  None. This is a bug fix that doesn't change any public API.

guide-linhnv:
Hi @mvanbeusekom,
Could you please help me review this PR?

PR #1783, opened 2026-06-08
docs: clarify that Position.heading carries course over ground (#1187)

aki1770-del:
Resolves #1187.

As discussed in #1187, `Position.heading` is named "heading" but actually carries **course over ground** (the direction of travel), not compass heading. In the issue, @mvanbeusekom proposed a documentation clarification rather than a breaking rename, and invited a PR.

This updates the dartdoc of `Position.heading` in `geolocator_platform_interface` to explain that the value is course over ground (distinct from compass heading), and that the `heading` name is retained for backward compatibility.

The documented native sources are accurate:
- Android: `geolocator_android/.../LocationMapper.java` maps `heading` from `Location.getBearing()`.
- iOS: `geolocator_apple/.../LocationMapper.m` maps `heading` from `CLLocation.course`.

This is **documentation-only** — no code, signature, or serialization changes; the `heading` field name and behaviour are unchanged, so it is non-breaking.

Per your request in the issue:
- Bumped `geolocator_platform_interface` `4.2.6` → `4.2.7`.
- Added a `## 4.2.7` CHANGELOG entry. I left the existing `## NEXT` `flutter_lints ^5.0.0` entry under `NEXT` so you can cut that on your own cadence — happy to fold it into this release instead if you'd prefer.

Verification (`geolocator_platform_interface`, Flutter 3.45 / Dart 3.11):
- `dart format` — clean (no changes)
- `flutter analyze` — No issues found!
- `flutter test` — All tests passed! (115/115)

Branched from `main` per your note that the repo no longer uses a `develop` branch.

Issue #1525, opened 2024-06-06
[Feature Request]: Expose Location.hasBearingAccuracy() from Android platform


JaseElder:
### Is there already an issue requesting this feature?

- [X] I have searched the [existing issues](https://github.com/baseflow/flutter-geolocator/issues).

### Please select affected platform(s)

- [X] Android
- [ ] iOS
- [ ] Linux
- [ ] macOS
- [ ] Web
- [ ] Windows

### Use case

It would be nice to have the Android `Location.hasBearingAccuracy()` call exposed to Flutter, as we were filtering out inaccurate heading readings using `Position.headingAccuracy`, but on some devices this always = 0.0 if `Location.hasBearingAccuracy()` returns `false`, so on those devices, we're throwing away all readings.

### Proposal

An extra field in `geolocator_platform_interface-4.2.2/lib/src/models/position.dart` to flag whether the devices has bearing/heading accuracy available.

TimHoogstrate:
Dear @JaseElder,

I'll label this as an enhancement, however, we will encourage you to propose a PR with the requested changes.

Kind regards,

PR #1793, opened 2026-08-21
feat: distinguish an unmeasured Position value from a measured 0.0

### LINKED ISSUE #1733: [Question]: Distinguish between real '0' and no-data '0'
### Please check the following before submitting a new issue.

- [x] I have searched the [existing issues](https://github.com/baseflow/flutter-geolocator/issues).
- [x] I have carefully [read the documentation](https://github.com/baseflow/flutter-geolocator/blob/main/geolocator/README.md).

### Please select for which platform(s) you need help

- [x] Android
- [ ] iOS
- [ ] Linux
- [ ] macOS
- [ ] Web
- [ ] Windows

### Your question

The following is extracted from within package implementation
```  
/// The speed at which the devices is traveling in meters per second over
  /// ground.
  ///
  /// The speed is not available on all devices. In these cases the value is
  /// 0.0.
  final double speed;

  /// The estimated speed accuracy of this position, in meters per second.
  ///
  /// The speedAccuracy is not available on all devices. In these cases the
  /// value is 0.0.
  final double speedAccuracy;
```

according to this we get `0`when there is no data. How do we distinguish whether this 0 means `no-data` or `real 0`?

Could `no-data` be changed to return `null`? Or expose getters to identify the value between `real` and `not-found`.

### Version

14.0.1

aki1770-del:
Resolves #1733 and #1525.

## The problem

`Position.fromMap` cannot distinguish a value the platform did not measure from the same value measured as `0.0`.

The Android side already gets this right. `LocationMapper.toHashMap` guards every optional field with the platform's own predicate and omits the key when it is false:

```java
if (location.hasAltitude()) position.put("altitude", location.getAltitude());
if (location.hasAccuracy()) position.put("accuracy", (double) location.getAccuracy());
if (location.hasBearing()) position.put("heading", (double) location.getBearing());
```

So the absence arrives at the platform interface correctly expressed, as a missing key. It is then discarded here:

```dart
static double _toDouble(dynamic value) {
  if (value == null) {
    return 0.0;
  }
  return value.toDouble();
}
```

The consequence is in both issues, in opposite directions:

- @Priyantha-Kingslake in #1733: *"according to this we get `0`when there is no data. How do we distinguish whether this 0 means `no-data` or `real 0`?"* — and, in the same thread, *"I have ~30% out of all location fixes with 0, 0 for heading and heading_accuracy"*.
- @JaseElder in #1525: *"we were filtering out inaccurate heading readings using `Position.headingAccuracy`, but on some devices this always = 0.0 if `Location.hasBearingAccuracy()` returns `false`, so on those devices, we're throwing away all readings."*

A consumer filtering `accuracy < threshold` accepts a position carrying no accuracy at all, because `0.0` passes every threshold. A consumer filtering `headingAccuracy > 0` discards every reading on devices that cannot report it. Both are reasonable code against the current contract.

For what it is worth: my own previous PR here, #1783, edited the doc comment on `Position.heading` and left the sentence *"In these cases the value is 0.0"* standing in text I had just rewritten, while both of these issues were already open. I did not see it then.

## What this changes

Adds seven `bool` fields to `Position`, one per optional value:

`hasAccuracy` · `hasAltitude` · `hasAltitudeAccuracy` · `hasHeading` · `hasHeadingAccuracy` · `hasSpeed` · `hasSpeedAccuracy`

`fromMap` sets each from what the map actually carried. `toJson` emits them, and `fromMap` prefers an explicit `has_*` key over key presence, so a round trip does not resurrect a measurement that never happened.

No platform code changes. The predicates are already being called on Android; this just stops throwing the answer away.

## Backwards compatibility

- Every new field is optional and defaults to `false`, following the existing `floor` / `isMocked` pattern in the same constructor.
- The reported numeric values are unchanged. Code that reads `accuracy`, `heading` or `speed` behaves exactly as before.
- `flutter test` for `geolocator_platform_interface`: **120 passing**, of which **115 are the pre-existing tests**, unchanged.

## Verification

Run against `geolocator_platform_interface` per the checks in `.github/workflows/geolocator_platform_interface.yaml`:

| command | result |
|---|---|
| `flutter pub get` | ok |
| `dart format --set-exit-if-changed .` | 39 files, 0 changed |
| `flutter analyze` | No issues found! |
| `flutter test --coverage` | +120: All tests passed |

The new test file was written first and run against the unmodified tree, where it fails to compile (`The getter 'hasAccuracy' isn't defined for the type 'Position'`). It passes 5/5 after the change.

Verified on a Flutter build that is not on the `stable` channel your CI pins, so please treat the local `dart format` result as indicative rather than CI-equivalent.

## What this does not do

- Only `geolocator_platform_interface` is changed. `geolocator_android`, `geolocator_apple`, `geolocator_linux`, `geolocator_web` and `geolocator_windows` are untouched and were not built or tested here.
- `geolocator_linux` in particular still coalesces GeoClue's nulls: `geoclue_x.dart` sets `altitudeAccuracy`, `headingAccuracy` and `speedAccuracy` to `0` unconditionally and applies `?? 0` to `altitude`, `heading` and `speed`, even though `GeoClueLocation` reports those as nullable. Wiring these flags there would be a separate change; I did not want to bundle it.
- The docs on the affected fields still describe the `0.0` behaviour. @TimHoogstrate noted in #1733 that the platform interface documentation needed updating; I have left that alone rather than mix a doc rewrite into this diff. Happy to follow up separately if useful.

Happy to split this differently, rename the fields, or drop the `toJson` change if you would rather keep the serialized shape fixed.

### DIFF
--- a/geolocator_platform_interface/CHANGELOG.md
+++ b/geolocator_platform_interface/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.3.0
+
+- Adds `hasAccuracy`, `hasAltitude`, `hasAltitudeAccuracy`, `hasHeading`, `hasHeadingAccuracy`, `hasSpeed` and `hasSpeedAccuracy` to `Position`, so a value the platform could not measure is distinguishable from the same value measured as `0.0`. The Android platform channel already omits a key when its `Location.hasX()` predicate is false; `Position.fromMap` substituted `0.0` for the missing value, and the distinction was lost. The reported numeric values are unchanged and every field defaults to `false`, so existing code is unaffected. Resolves [#1733](https://github.com/Baseflow/flutter-geolocator/issues/1733) and [#1525](https://github.com/Baseflow/flutter-geolocator/issues/1525).
+
 ## 4.2.8
 
 - Updates dependencies to their latest versions.
--- a/geolocator_platform_interface/lib/src/models/position.dart
+++ b/geolocator_platform_interface/lib/src/models/position.dart
@@ -19,6 +19,13 @@ class Position {
     required this.speedAccuracy,
     this.floor,
     this.isMocked = false,
+    this.hasAccuracy = false,
+    this.hasAltitude = false,
+    this.hasAltitudeAccuracy = false,
+    this.hasHeading = false,
+    this.hasHeadingAccuracy = false,
+    this.hasSpeed = false,
+    this.hasSpeedAccuracy = false,
   });
 
   /// The latitude of this position in degrees normalized to the interval -90.0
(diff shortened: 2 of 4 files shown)


mvanbeusekom (approved):
I like this solution over others that introduce nullable fields, as this will not introduce a breaking change. 

The code looks good, thank you for this contribution. I will go ahead and merge the PR and publish the new version.

